PR #4, opened 2015-06-15
Win build

kkirstein:
- add symbol definitions for Windows build
- fix function signature for Windows in remutex package
- add build instruction for Windows to Readme

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,11 @@ Cargo.lock
 **/target/release
 /tmp.*
 **/*.h5
+
+# vim
+*~
+*.swp
+
+# vagrant
+.vagrant/
+
--- a/README.md
+++ b/README.md
@@ -7,6 +7,21 @@ Thread-safe Rust bindings and high-level wrappers for the HDF5 library API.
 Note that this project is in its early development stage and hence things are likely to change
 and break on a regular basis.
 
+## Building
+
+### Windows
+
+Building hdf5-rs on Windows currently needs some manual preparation steps:
+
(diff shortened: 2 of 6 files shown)


kkirstein:
@aldanor I just created a pull request with the Windows-enabled version. I added some build instructions to the Readme. I also had a short look at appveyor, looks like a good option, although I have no experience with CI from the cloud.
As I will be on vacation for the rest of the week, please don't expect any answers from me during the next days, I will take a closer look at appveyor next week.

aldanor:
@kkirstein 

Thanks a lot of the PR! I won't have time to look at it in detail till weekend but it looks good from the first glance -- although I have no way of testing it actually works on Windows right now other than taking your word on it -- would be really great to try and set up a CI build for it.

I have no experience with Appveyor but it looks reasonable. Maybe you could take a look at PyTables build? See PR here: https://github.com/PyTables/PyTables/pull/426

As for MSVC interoperability, it looks like there's some very recent ongoing work going upstream, see here for more details:
- https://github.com/rust-lang/rfcs/issues/1061
- https://github.com/rust-lang/rust/pull/25350

Ideally, we'd want to support both mingw and msvc builds, if possible at all.

kkirstein:
@aldanor Thanks, for pointing to the discussions on rust-lang!

aldanor:
@kkirstein You're welcome! By the way, I've noticed that many rust-lang repos were set up for appveyor builds about a week ago -- see, for instance, https://github.com/rust-lang/regex/commit/1be2feedb6203ce6bd430547152f11611fa56c87 -- maybe that could be a good starting point.

Good call with `#[inline]`, too.

aldanor:
Closed in favor of #7.

PR #5, opened 2015-06-16
Use winapi/kernel32 for Windows FFI

retep998:
Also adds `#[inline]` to all the methods because they're just thin wrappers anyway.
Blocked on landing new versions of `winapi` and `kernel32` on crates.io.

### DIFF
--- a/remutex/Cargo.toml
+++ b/remutex/Cargo.toml
@@ -5,3 +5,5 @@ version = "0.0.1"
 
 [dependencies]
 libc = "*"
+winapi = "0.1"
+kernel32-sys = "0.1"
--- a/remutex/src/windows.rs
+++ b/remutex/src/windows.rs
@@ -6,60 +6,45 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
+extern crate winapi;
+extern crate kernel32;
+
 use std::mem;
 use std::cell::UnsafeCell;
 
-pub struct ReentrantMutex { inner: UnsafeCell<ffi::CRITICAL_SECTION> }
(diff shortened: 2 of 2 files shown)


aldanor:
@retep998 Thanks for the PR! This would be more @kkirstein's cup of tea, I guess, since I currently have no way of testing anything on win32. It's not really obvious to me though what is the benefit for us for doing this since we'll have to pull two extra crates as dependencies to achieve exactly the same thing and link to the same API, would you care to elaborate?

The whole `remutex` crate is a direct port of the core `std::sync` functionality (buried somewhere in `sys_common`) which is already available in nightly Rust. It's unclear if it's ever going to be made public though...

retep998:
`winapi` and friends are used by lots of projects, and don't take long to build (on non-windows platforms they're no-ops entirely). I write PRs like this mostly out of my own selfish desire to make the world depend on me, but also to make sure that people don't use incorrect definitions like `pub type ULONG_PTR = c_ulong;` because `ULONG_PTR` is actually 64 bits on 64-bit platforms while `c_ulong` is only 32 bits.

aldanor:
@retep998 I see, thanks for explaining! I guess I'll just stick with port of the libstd implementation for now though and see if it lands in 1.1 stable (in which case we'll just remove the local crate and use std) but otherwise we'll sure keep an eye on `winapi` :)

Cheers,
-I

kkirstein:
@retep998 Thanks for pointing to `winapi` and friends. At the moment I agree with @aldanor not to add further dependencies just for the `remutex` package. Adding `#inline` to the wrappers is a good idea anyway...
I will definitely keep an eye on your crates and we might rethink our dependencies, when it is more clear what Rust's std-libs are covering and what not.

PR #7, opened 2015-07-06
Appveyor setup

kkirstein:
@aldanor I somehow messed up my branches, but now you should see the appveyor setup as a single commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,11 @@ Cargo.lock
 **/target/release
 /tmp.*
 **/*.h5
+
+# vim
+*~
+*.swp
+
+# vagrant
+.vagrant/
+
--- a/README.md
+++ b/README.md
@@ -7,6 +7,21 @@ Thread-safe Rust bindings and high-level wrappers for the HDF5 library API.
 Note that this project is in its early development stage and hence things are likely to change
 and break on a regular basis.
 
+## Building
+
+### Windows
+
+Building hdf5-rs on Windows currently needs some manual preparation steps:
+
+* A HDF5 binary, namely the shared library `hdf5.dll` is needed for linking. For (currently) unknown reasons, the prebuilt binaries from [HDF-Group](http://www.hdfgroup.org/) do not work. It has to be build with gcc. Instructions for building HDF5 on Windows can be found [here](http://www.hdfgroup.org/HDF5/release/cmakebuild.html). For building the [TDM distribution](http://tdm-gcc.tdragon.net/)of MinGW-GCC is recommended, as it contains bintools for both 32bit & 64bit.
+
+* Set the environment variable `HDF5_LIBDIR` to point to the folder with the newly build `hdf5.dll`. Explanation: `pkg-config` will silently fail if not present and the path from the before-mentioned environment variable is added to the rustc commands by cargo. (Hint: Avoid path names with spaces, as they are difficult to escape correctly).
+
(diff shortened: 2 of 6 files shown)


aldanor:
@kkirstein I wonder if we can use a prebuilt binary from continuum so we don't have to link to `hdf5.dll` in your repo? I think they build it via mingw.

So it could be something like this (7z is already available on appveyor):

```yaml
- ps: Start-FileDownload https://repo.continuum.io/pkgs/free/win-64/hdf5-1.8.15.1-2.tar.bz2
- 7z x hdf5-1.8.15.1-2.tar.bz2
- 7z x hdf5-1.8.15.1-2.tar
```

aldanor:
Also, it would be nice (but is no biggie) if this whole PR was squashed into just one commit -- you can just do a `git rebase --interactive`, mark the commits to squash, and then `git push --force`.

kkirstein:
@aldanor currently I have to use a hdf5.dll specially compiled with gcc (instead of msvc). Originally I wanted to pull that in from dropbox, but it looks like ```Start-FileDownload``` has some problems with dropbox html-redirections. Do you know, how the binaries from continuum are compiled?

kkirstein:
@aldanor Regarding the squashed commit: This is exactly what I (tried) to do, but I somehow messed it up.

aldanor:
@kkirstein re: rebasing, would you care to try it again? :)

Re: `hdf5.dll`, it looks like continuum is also using the MSVC stack to build hdf5.dll... it also looks like MinGW is not officially supported by HDF5 Group at all... 

By the way, have you tried bulding / running the tests with MSVC hdf5.dll on Rust nightly? If it works, we can use the official HDF5 binaries and only run the tests on nightly till MSVC support lands in Rust 1.2 stable?

kkirstein:
@aldanor I just tried the msvc target of rust-nightly, but it seems to be still quite alpha (missing symbols from msvc-runtime). I didn't think of using "rust-nightly-gnu" with the MSVC-hdf5.dll and will give it a try...

kkirstein:
@aldanor Well actually rebasing for appveyor setup was successful, the problem is that my branch contains other changes as well (Yes I know, it is not a good idea to perform several changes within one branch..). Maybe you just want to ```cherry-pick``` the appveyor commit: https://github.com/aldanor/hdf5-rs/commit/c13560a5e365dc036502c3ca264496fc7cf78a4d
Otherwise, I just re-fork your ```master``` and generate another PR. (Sorry for the mess, just learning how to collaborate with GitHub).

kkirstein:
@aldanor BTW, ```rustc-nightly-xx-gnu``` does not work with MSVC hdf5. So, we have to stick with the MinGW-compiled hdf5 for now. Although, this is not officially supported by HDFGroup (anymore), it compiles out-of-the box, thanks to ```cmake```.

aldanor:
Closed via 67cc73a628c1bc0d1b3feea4a098b86e137a01f0

aldanor:
Thanks for setting this up, @kkirstein! It seems to work just fine (aside from the whole mingw business), now we just have to wait for MSVC support to land in Rust (hopefully in 1.2.0?), then we can switch to using only the official binaries from HDF Group.

kkirstein:
@aldanor In the meantime I've also setup a MSVC build on appveyor with rust-nightly, but it still fails compiling basic modules, e.g. pkg-config. So, I would suggest to wait for a more stable MSVC support.

PR #20, opened 2019-01-04
read_slice() / write_slice() methods to permit dataset slice IO

pmarks:
Add a reader method that supports fetching ndarray slices out of a dataset, letting the user avoid reading the entire dataset. Any feedback on the general approach?  If it seems like we're on the right track here I'll 1-d and 2-d special cases and writer support.

TODOs:
 - [x] support SliceOrIndex::Index() case
 - [x] special case 1D and 2D sliced reads
 - [x] writer support

### DIFF
--- a/hdf5-rs/src/hl/container.rs
+++ b/hdf5-rs/src/hl/container.rs
@@ -2,7 +2,8 @@ use std::fmt::{self, Debug};
 use std::mem;
 use std::ops::Deref;
 
-use ndarray::{Array, Array1, Array2, ArrayD, ArrayView, ArrayView1};
+use ndarray::{Array, Array1, Array2, ArrayD, ArrayView, ArrayView1, Ix1, Ix2};
+use ndarray::{SliceInfo, SliceOrIndex};
 
 use libhdf5_sys::h5a::{H5Aget_space, H5Aget_storage_size, H5Aget_type, H5Aread, H5Awrite};
 use libhdf5_sys::h5d::{H5Dget_space, H5Dget_storage_size, H5Dget_type, H5Dread, H5Dwrite};
@@ -35,17 +36,87 @@ impl<'a> Reader<'a> {
         self
     }
 
-    fn read_into_buf<T: H5Type>(&self, buf: *mut T) -> Result<()> {
+    fn read_into_buf<T: H5Type>(
+        &self, buf: *mut T, fspace: Option<&Dataspace>, mspace: Option<&Dataspace>,
+    ) -> Result<()> {
         let file_dtype = self.obj.dtype()?;
         let mem_dtype = Datatype::from_type::<T>()?;
         file_dtype.ensure_convertible(&mem_dtype, self.conv)?;
         let (obj_id, tp_id) = (self.obj.id(), mem_dtype.id());
+
+        let fspace_id = fspace.map_or(H5S_ALL, |f| f.id());
+        let mspace_id = mspace.map_or(H5S_ALL, |m| m.id());
+
         if self.obj.is_attr() {
             h5try!(H5Aread(obj_id, tp_id, buf as *mut _));
         } else {
-            h5try!(H5Dread(obj_id, tp_id, H5S_ALL, H5S_ALL, H5P_DEFAULT, buf as *mut _));
+            h5try!(H5Dread(obj_id, tp_id, mspace_id, fspace_id, H5P_DEFAULT, buf as *mut _));
         }
         Ok(())
+    }
+
+    /// Reads a slice of an n-dimensional array.
+    /// If the array has a fixed number of dimensions, it must match the dimensionality of
+    /// dataset. Use the multi-dimensional slice macro `s![]` from `ndarray` to conveniently create
+    /// a multidimensional slice.
+    pub fn read_slice<T, S, D>(&self, slice: &SliceInfo<S, D>) -> Result<Array<T, D>>
+    where
+        T: H5Type,
+        S: AsRef<[SliceOrIndex]>,
+        D: ndarray::Dimension,
+    {
+        ensure!(!self.obj.is_attr(), "slicing cannot be used on attribute datasets");
+
+        let shape = self.obj.get_shape()?;
+        if let Some(ndim) = D::NDIM {
+            let obj_ndim = shape.ndim();
+            ensure!(obj_ndim == ndim, "ndim mismatch: expected {}, got {}", ndim, obj_ndim);
+        }
+
+        let slice_s: &[SliceOrIndex] = slice.as_ref();
+        let slice_dim = slice_s.len();
+        if shape.ndim() != slice_dim {
+            let obj_ndim = shape.ndim();
+            ensure!(
+                obj_ndim == slice_dim,
+                "slice dimension mismatch: dataset has {} dims, slice has {} dims",
+                obj_ndim,
+                slice_dim
+            );
+        }
+
+        if shape.ndim() == 0 {
+            // Fall back to a simple read for the scalar case
+            // Slicing has no effect
+            self.read()
+        } else {
+            let fspace = self.obj.space()?;
+            let out_shape = fspace.select_slice(slice)?;
+
+            // Remove dimensions from out_shape that were an
+            //SliceOrIndex::Index in the slice
+            let reduced_shape: Vec<usize> = slice_s
+                .iter()
+                .zip(out_shape.iter().cloned())
+                .filter(|(slc, _)| match slc {
+                    SliceOrIndex::Index(_) => false,
+                    _ => true,
+                })
+                .map(|(_, sz)| sz)
+                .collect();
+
+            let mspace = Dataspace::try_new(&out_shape, false)?;
+            let size = out_shape.iter().product();
+            let mut vec = Vec::with_capacity(size);
+            unsafe {
+                vec.set_len(size);
+            }
+
+            self.read_into_buf(vec.as_mut_ptr(), Some(&fspace), Some(&mspace))?;
+
+            let arr = ArrayD::from_shape_vec(reduced_shape, vec).str_err()?;
+            arr.into_dimensionality().str_err()
+        }
     }
 
     /// Reads a dataset/attribute into an n-dimensional array.
@@ -70,7 +141,7 @@ impl<'a> Reader<'a> {
         unsafe {
             vec.set_len(size);
         }
-        self.read_into_buf(vec.as_mut_ptr()).map(|_| vec)
+        self.read_into_buf(vec.as_mut_ptr(), None, None).map(|_| vec)
     }
 
     /// Reads a dataset/attribute into a 1-dimensional array.
@@ -80,11 +151,33 @@ impl<'a> Reader<'a> {
         self.read()
     }
 
+    /// Reads the given `slice` of the dataset into a 1-dimensional array.
+    ///
+    /// The dataset must be 1-dimensional.
+    pub fn read_slice_1d<T, S>(&self, slice: &SliceInfo<S, Ix1>) -> Result<Array1<T>>
+    where
+        T: H5Type,
+        S: AsRef<[SliceOrIndex]>,
+    {
+        self.read_slice(slice)
+    }
+
     /// Reads a dataset/attribute into a 2-dimensional array.
     ///
     /// The dataset/attribute must be 2-dimensional.
     pub fn read_2d<T: H5Type>(&self) -> Result<Array2<T>> {
         self.read()
+    }
+
+    /// Reads the given `slice` of the dataset into a 2-dimensional array.
+    ///
+    /// The dataset must be 2-dimensional.
+    pub fn read_slice_2d<T, S>(&self, slice: &SliceInfo<S, Ix2>) -> Result<Array2<T>>
+    where
+        T: H5Type,
+        S: AsRef<[SliceOrIndex]>,
+    {
+        self.read_slice(slice)
     }
 
     /// Reads a dataset/attribute into an array with dynamic number of dimensions.
@@ -97,7 +190,7 @@ impl<'a> Reader<'a> {
         let obj_ndim = self.obj.get_shape()?.ndim();
         ensure!(obj_ndim == 0, "ndim mismatch: expected scalar, got {}", obj_ndim);
         let mut val: T = unsafe { mem::uninitialized() };
-        self.read_into_buf(&mut val as *mut _).map(|_| val)
+        self.read_into_buf(&mut val as *mut _, None, None).map(|_| val)
     }
 }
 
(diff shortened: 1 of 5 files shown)


hdf5-rs/src/hl/space.rs
```diff
@@ -49,6 +54,69 @@ impl Deref for Dataspace {
...
+    }
+
+    /// Select a slice (known as a 'hyperslab' in HDF5 terminology) of the Dataspace.
+    /// Returns the shape of array that is capable of holding the resulting slice.
+    /// Useful when you want to read a subset of a dataset.
+    pub fn select_slice<T, D>(&self, slice: SliceInfo<T,D>) -> Result<Vec<Ix>> where
+    T: AsRef<[SliceOrIndex]>, D: ndarray::Dimension
+    {
+        let shape = self.dims();
+        let mut start_vec = Vec::new();
```
Enet4: These vectors can be allocated with the expected capacity.

```suggestion
        let mut start_vec = Vec::with_capacity(ss.len());
```

hdf5-rs/tests/common/gen.rs
```diff
@@ -17,6 +17,55 @@ pub fn gen_ascii<R: Rng + ?Sized>(rng: &mut R, len: usize) -> String {
     iter::repeat(()).map(|_| rng.sample(Alphanumeric)).take(len).collect()
 }
 
+/// Generate a random slice of elements inside the given `shape` dimension.
+pub fn gen_slice<R: Rng + ?Sized>(rng: &mut R, shape: &[usize]) -> Vec<SliceOrIndex> {
+
+    let mut r = Vec::new();
```
Enet4: `Iterator` all the things:

```rust
shape.into_iter()
    .map(|s| gen_slice_one_dim(rng, *s))
    .collect()
```

hdf5-rs/src/hl/container.rs
```diff
@@ -35,19 +36,56 @@ impl<'a> Reader<'a> {
...
         }
         Ok(())
     }
 
+    /// Reads a slice of an n-dimensional array.
+    /// If the array has a fixed number of dimensions, it must match the dimensionality of
+    /// dataset.  FIXME - ensure this: Sliced reads are not allowed from attributes.
+    /// Use the multidimsional slice macro `s![]` from `ndarray` to conveniently create
+    /// a multidimensional slice.
+    pub fn read_slice<T,S,D>(&self, slice: SliceInfo<S,D>) -> Result<Array<T, D>>
```
aldanor: Shouldn't the slice be a ref?

Looking at [`ArrayBase::slice()`](https://docs.rs/ndarray/0.12.1/ndarray/struct.ArrayBase.html#method.slice), it takes a `&SliceInfo<...>`?

hdf5-rs/src/hl/container.rs
```diff
@@ -35,19 +36,56 @@ impl<'a> Reader<'a> {
...
+        }
+        if shape.ndim() == 0 {
+            // Fall back to a simple read for the scalar case
+            // Slicing has no effect
+            self.read()
+        }  else {
+            let fspace = self.obj.space()?;
+            let out_shape = fspace.select_slice(slice)?;
+            let mspace = Dataspace::create_simple(&out_shape)?;
+            let size = out_shape.iter().fold(1 ,|p, v| p * (*v));
```
aldanor: There's `iter().product()` (sometimes the type inference fails and you will have to specify its return type via turbofish).

hdf5-rs/src/hl/container.rs
```diff
@@ -35,19 +36,56 @@ impl<'a> Reader<'a> {
...
+
+        let shape = self.obj.get_shape()?;
+        if let Some(ndim) = D::NDIM {
+            let obj_ndim = shape.ndim();
+            ensure!(obj_ndim == ndim, "ndim mismatch: expected {}, got {}", ndim, obj_ndim);
+        }
+        if shape.ndim() == 0 {
+            // Fall back to a simple read for the scalar case
+            // Slicing has no effect
+            self.read()
```
aldanor: It's an edge case, but - should this fail loudly? E.g. it would fail in h5py, I think.

hdf5-rs/src/hl/container.rs
```diff
@@ -35,19 +36,56 @@ impl<'a> Reader<'a> {
...
+            let mspace = Dataspace::create_simple(&out_shape)?;
+            let size = out_shape.iter().fold(1 ,|p, v| p * (*v));
+            let mut vec = Vec::with_capacity(size);
+            unsafe {
+                vec.set_len(size);
+            }
+
+            self.read_into_buf(vec.as_mut_ptr(), Some((&fspace, &mspace))).unwrap();
+
+            let arr = ArrayD::from_shape_vec(out_shape, vec).str_err().unwrap();
```
aldanor: No unwraps please?

hdf5-rs/src/hl/container.rs
```diff
@@ -35,19 +36,56 @@ impl<'a> Reader<'a> {
...
+            let fspace = self.obj.space()?;
+            let out_shape = fspace.select_slice(slice)?;
+            let mspace = Dataspace::create_simple(&out_shape)?;
+            let size = out_shape.iter().fold(1 ,|p, v| p * (*v));
+            let mut vec = Vec::with_capacity(size);
+            unsafe {
+                vec.set_len(size);
+            }
+
+            self.read_into_buf(vec.as_mut_ptr(), Some((&fspace, &mspace))).unwrap();
```
aldanor: No unwraps?

hdf5-rs/src/hl/space.rs
```diff
@@ -1,9 +1,14 @@
 use std::fmt::{self, Debug};
 use std::ops::Deref;
 use std::ptr;
+use ndarray::{SliceInfo, SliceOrIndex};
```
aldanor: The imports in this crate are usually split like this:

```rust
use std::foo;
use std::bar;

use third_party1::foo;
use third_party2::bar;

use hdf5_rs_workspace_stuff;

use crate::local::stuff;
```

hdf5-rs/src/hl/space.rs
```diff
@@ -49,6 +54,69 @@ impl Deref for Dataspace {
 }
 
 impl Dataspace {
+
```
aldanor: Extra newline

hdf5-rs/src/hl/space.rs
```diff
@@ -49,6 +54,69 @@ impl Deref for Dataspace {
...
+    /// Create a simple dataspace covering the given `shape`
+    pub fn create_simple(shape: &[Ix]) -> Result<Dataspace> {
```
aldanor: How is this different from `Dataspace::try_new()?` (aside from a better name perhaps) Or should it at least reuse that?

(dataspaces/shapes/dims will likely be heavily refactored in the future anyway to integrate better with ndarray)

aldanor: `Result<Dataspace>` -> `Result<Self>`, it's slightly nicer / more flexible.

hdf5-rs/src/hl/space.rs
```diff
@@ -49,6 +54,69 @@ impl Deref for Dataspace {
...
+        let shape_u64: Vec<u64> = shape.iter().map(|v| *v as u64).collect();
+        let id = h5try!(H5Screate_simple(shape.len() as i32, shape_u64.as_ptr(), ptr::null()));
+        Ok(Dataspace::from_handle(Handle::try_new(id)?))
+    }
+
+    /// Select a slice (known as a 'hyperslab' in HDF5 terminology) of the Dataspace.
+    /// Returns the shape of array that is capable of holding the resulting slice.
+    /// Useful when you want to read a subset of a dataset.
+    pub fn select_slice<T, D>(&self, slice: SliceInfo<T,D>) -> Result<Vec<Ix>> where
```
aldanor: `slice` can be taken by ref?

hdf5-rs/src/hl/space.rs
```diff
@@ -49,6 +54,69 @@ impl Deref for Dataspace {
...
+        h5try!(H5Sselect_hyperslab(self.id(), H5S_SELECT_SET, start_vec.as_ptr(), stride_vec.as_ptr(), count_vec.as_ptr(), ptr::null()));
+        Ok(shape_vec)
+    }
+     
+    fn get_start_stride_count(v: &SliceOrIndex, shape: &Ix) -> (u64, u64, u64) {
+        match v {
+            SliceOrIndex::Slice { start, end, step } => {
+                let end = end.unwrap_or(*shape as isize);
+                if end > (*shape as isize) {
+                    panic!("invalid selection");
```
aldanor: No panics please - this should return a `Result<>`

hdf5-rs/src/hl/space.rs
```diff
@@ -49,6 +54,69 @@ impl Deref for Dataspace {
...
+                }
+
+                if *step < 1 {
+                    panic!("Step must be >= 1 (got {})", step);
```
aldanor: No panics

aldanor:
To leave some comments on the general approach I'd need to read some official HDF5 docs on hyperslabs, will get back later with that :)

hdf5-rs/src/hl/container.rs
```diff
@@ -35,19 +36,56 @@ impl<'a> Reader<'a> {
         self
     }
 
-    fn read_into_buf<T: H5Type>(&self, buf: *mut T) -> Result<()> {
+    fn read_into_buf<T: H5Type>(&self, buf: *mut T, file_space: Option<(&Dataspace, &Dataspace)>) -> Result<()> {
```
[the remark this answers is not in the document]
pmarks: agreed

hdf5-rs/src/hl/space.rs
pmarks: sorry, missed try_new. Got rid of this method.

hdf5-rs/src/hl/container.rs
pmarks: I think this is consistent with h5py. There's is only one 0-dimensional slice (the empty array), and when applied to a scalar dataset it just returns the value:

```
$ipython
In [13]: f = h5py.File("test.h5", "w")

In [14]: f.create_dataset("t1", (), data=5)
Out[14]: <HDF5 dataset "t1": shape (), type "<i8">

In [15]: tt = f["t1"][()]

In [16]: tt
Out[16]: 5
```

pmarks:
@aldanor @Enet4  Thanks for the quick reviews! I've made all the suggested changes expect for the point about slicing a scalar dataset.

Re the selelct_hyperslab pattern: from past work I have the impression that this is a very common style.  This setup is very similar to how h5py does things. Here's a C example: https://support.hdfgroup.org/ftp/HDF5/current/src/unpacked/examples/h5_subset.c

I'll keep going with this general style unless there are additional style ideas.

hdf5-rs/src/hl/container.rs
aldanor: Yea but your slice dimensions could be anything here? (because you're checking `obj.shape()`)

If you try this in h5py, it should fail:

```python
>>> f['t1'][1:2]
```

hdf5-rs/src/hl/space.rs
```diff
@@ -49,7 +53,66 @@ impl Deref for Dataspace {
...
+        ));
+        Ok(shape_vec)
+    }
+
+    fn get_start_stride_count(v: &SliceOrIndex, shape: &Ix) -> Result<(u64, u64, u64)> {
+        match v {
+            SliceOrIndex::Slice { start, end, step } => {
+                let end = end.unwrap_or(*shape as isize);
+                if end > (*shape as isize) {
+                    return Err(Error::Internal("slice is larger than dataset".to_string()));
```
aldanor: You could just do `fail!("slice is larger than the dataset")` (or "slice extends beyond dataspace bounds" since this is about dataspace).

aldanor: In fact you can even do an `ensure!(end <= shape as _,"slice extends beyond dataspace bounds")`.

hdf5-rs/src/hl/space.rs
```diff
@@ -49,7 +53,66 @@ impl Deref for Dataspace {
...
+            H5S_SELECT_SET,
+            start_vec.as_ptr(),
+            stride_vec.as_ptr(),
+            count_vec.as_ptr(),
+            ptr::null()
+        ));
+        Ok(shape_vec)
+    }
+
+    fn get_start_stride_count(v: &SliceOrIndex, shape: &Ix) -> Result<(u64, u64, u64)> {
```
aldanor: No point in taking `&Ix` by reference here since it's a scalar, can just do an `Ix`.

Also, `shape` is a misleading name here? Maybe `len` would have ben better?

hdf5-rs/src/hl/space.rs
```diff
@@ -49,7 +53,66 @@ impl Deref for Dataspace {
...
+        match v {
+            SliceOrIndex::Slice { start, end, step } => {
+                let end = end.unwrap_or(*shape as isize);
+                if end > (*shape as isize) {
+                    return Err(Error::Internal("slice is larger than dataset".to_string()));
+                }
+
+                if *step < 1 {
```
aldanor: `ensure!(*step >= 1, "step must be >= 1 (got {})", step)`

hdf5-rs/src/hl/space.rs
```diff
@@ -49,7 +53,66 @@ impl Deref for Dataspace {
...
+                    return Err(Error::Internal(format!("Step must be >= 1 (got {})", step)));
+                }
+
+                if end < *start {
```
aldanor: This whole thing (including logic below) probably simplifies to something like this?

```rust
1 + (end - start - 1).max(0) / step
```

The corner case is when `end == start` -- wondering why do you return `count=1` when `end < start`, but `count=0` when `end == start`?
```

Comment (pmarks) [reply hdf5-rs/src/hl/space.rs]: > ```rust
> 1 + (end - start - 1).max(0) / step
> ```
This will give `count = 1` when `end == start`, which I don't want.

> The corner case is when `end == start` -- wondering why do you return `count=1` when `end < start`, but `count=0` when `end == start`?

`end < start` is handled above. I changed the comparison to emphasize that `count=0` for empty intervals.  `1 + (end - start - 1).max(0) / step`

File: hdf5-rs/src/hl/space.rs (hunk 17 above)
Comment (pmarks) [reply hdf5-rs/src/hl/space.rs]: yes, thanks

File: hdf5-rs/src/hl/space.rs (hunk 16 above)
Comment (pmarks) [reply hdf5-rs/src/hl/space.rs]: good call, fixed.

File: hdf5-rs/src/hl/space.rs (hunk 15 above)
Comment (pmarks) [reply hdf5-rs/src/hl/space.rs]: yes, done.

**Comment by pmarks** [comment]
@aldanor OK, I've added 1-d and 2-d readers, a general writer, added some additional checks, and generalized it to support 'single index' `SliceOrIndex::Index` slices, which cause the dimensionality of the output of the slice to change. Tests are passing on stable but not nightly, which looks like a compiler bug to me.

Anything else you want me to tackle here?

File: hdf5-rs/src/hl/container.rs [hunk 19]
```diff
@@ -91,6 +92,19 @@ impl<'a> Reader<'a> {
         } else {
             let fspace = self.obj.space()?;
             let out_shape = fspace.select_slice(slice)?;
+
+            // Remove dimensions from out_shape that were an
+            //SliceOrIndex::Index in the slice
+            let reduced_shape: Vec<usize> = slice_s
+                .iter()
+                .zip(out_shape.iter().cloned())
+                .filter(|(slc, sz)| match slc {
```
aldanor: Minor: I think this filter + map could be combined:

```rust
.filter_map(|(slc, sz)| match slc {
    SliceOrIndex::Index(_) => None,
    _ => Some(sz)
})
```

hdf5-rs/src/hl/container.rs
```diff
@@ -261,18 +277,41 @@ impl<'a> Writer<'a> {
...
+                        data_shape_hydrated.push(data_shape[pos]);
+                        pos += 1;
+                    }
+                }
+            }
+
+            let mspace = Dataspace::try_new(&slice_shape, false)?;
 
             // FIXME - we can handle non-standard input arrays by creating a memory space
             // that reflects the same slicing/ordering that this ArrayView represents.
```
aldanor: Wondering - how hard would it be to support non standard layouts? As long as it's convertible to a hyperslab spec, it should be fine, no?

I'm thinking maybe all this "hyperslab" <--> "ndarray indexing" business should be moved out into a separate API to simplify all this logic here (and tested separately, without even touching HDF5 stuff). Not saying it should be done in this PR, just generally/later. Thoughts?

hdf5-rs/tests/test_dataset.rs
```diff
@@ -67,6 +67,9 @@ where
         let sliced_dataset = arr.slice(slice.as_ref());
 
         // Ensure that the H5 sliced read matches the ndarray slice of the original array.
+        if sliced_read != sliced_dataset {
+            println!("{:?}", slice);
```
aldanor: `assert_eq!()`?

hdf5-rs/src/hl/container.rs
```diff
@@ -223,11 +239,9 @@ impl<'a> Writer<'a> {
         Ok(())
     }
 
-
     /// Reads a slice of an n-dimensional array.
```
aldanor: "Writes"? Should probably be also clarified what exactly is being sliced here.

hdf5-rs/src/hl/container.rs
```diff
@@ -261,18 +277,41 @@ impl<'a> Writer<'a> {
...
             let view = arr.into();
+            let data_shape = view.shape();
+
+            // Restore dimensions that are SliceOrIndex::Index in the slice.
+            let mut data_shape_hydrated = Vec::new();
+            let mut pos = 0;
+            for s in slice_s {
+                match s {
+                    SliceOrIndex::Index(_) => {
+                        data_shape_hydrated.push(1);
```
aldanor: 1 or 0?

hdf5-rs/src/hl/container.rs
pmarks: yes, definitely - should be doable, and should be abstracted out of this code. Filed as #21.

hdf5-rs/src/hl/container.rs
pmarks: 1 -- SliceOrIndex::Index take a single element from that dimension.

aldanor:
@pmarks I've updated the feature/2018 branch with some of the recent changes (including fixes to the CI scripts so hopefully random failures will go away) -- could you please rebase your branch?

Maybe also squash most/all of your commits as well as you do it?

hdf5-rs/src/hl/container.rs
pmarks: you're correct -- this is now by the `if shape.ndim() != slice_dim {` check above.

pmarks:
Ok, everything is rebased & squashed. fmt clean & no warnings.  Thanks!

aldanor:
@pmarks Thanks!

(There's stuff to clean up and refactor, but we'll figure it later)

PR #22, opened 2019-01-10
[WIP] feature flag to statically link hdf5 from conda binaries

pmarks:
This is convenient way to quickly get static binaries, without having to think about where HDF5 is coming from & how to deploy the dll/so/dylibs.  

A better name for the feature is possibly 'conda-static'?  Wondering what people think about an option like this. 

I'm also doing something similar here:
https://github.com/pmarks/rust-intel-mkl/blob/master/build.rs
which is working nicely. Originally modeled on the tensorflow bindings:
https://github.com/tensorflow/rust/blob/master/tensorflow-sys/build.rs

TODO:
- [ ] test the `conda` feature in Travis

### DIFF
--- a/hdf5-rs/Cargo.toml
+++ b/hdf5-rs/Cargo.toml
@@ -35,3 +35,7 @@ tempdir = "0.3"
 
 [build-dependencies]
 libhdf5-lib = { path = "../libhdf5-lib", version = "0.2.0" }
+
+[features]
+# Use static HDF5 libs downloaded from conda.
+conda = ["libhdf5-lib/conda"]
--- a/libhdf5-lib/Cargo.toml
+++ b/libhdf5-lib/Cargo.toml
@@ -10,5 +10,16 @@ homepage = "https://github.com/aldanor/hdf5-rs"
 description = "Build script support for the HDF5 library."
 edition = "2018"
 
+links = "hdf5"
+
 [build-dependencies]
 pkg-config = "0.3"
+rust-crypto = "0.2"
+curl = "*"
+bzip2 = "0.3.3"
+tar = "*"
+
+[features]
+# Use static HDF5 libs downloaded from conda.
+conda = []
+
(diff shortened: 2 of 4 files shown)


aldanor:
Hi @pmarks, this is pretty neat idea, especially given that the large portion of target audience for this crate would be already using conda in this way or another. I'd probably exclude conda from feature name though so as to not to scare away people who have no idea what conda is. Maybe "download-18-static" or "download-110-static" (see below). Or "download-hdf5-18" or something?

Here's a few random thoughts:
- Hard-coding conda package names and their hashes seems a bit fragile. I'd rather prefer to recover the latest versions from `repodata.json` available in the repo (which I'd probably parse with regex and not serde since its schema is not guaranteed and sometimes they add new fields).
- Why HDF5 1.8 only? Most package managers like homebrew have been shipping 1.10 as a default (like if you do brew install hdf5) for quite a while now, with 1.8 being an opt-in (brew install hdf5@1.8). So maybe separate -18 and -110 options? Do we need more fine-grained versioning?
- Just a note of warning: conda will change package format at some point (maybe later this year) to .deb-style packages, see [this preliminary doc](https://docs.google.com/document/d/1HGKsbg_j69rKXPihhpCb1kNQSE8Iy3yOsUU2x68x8uw/edit?usp=sharing). 
- This method installs static libraries, which is nice in some cases. But if we're talking about conda, maybe we should also support another use case then: you're building a relocatable conda-package (via conda-build) and want to include your rust binary/library in it. In order for it to be relocatable, Rust's build system would need to know how to find libraries in conda environments based on the prefix, on different platforms. Then you could link dynamically to conda-installed hdf5 in the current environment (and thus use conda-build and it would magically "just work"). Would this be useful?
- Re: zlib - do we have to link to it, statically? (I don't quite remember why)

aldanor:
Hmm I'm not sure why it shows that I closed this PR although I don't remember doing so.

That being said, let me get back to it after 0.3 release -- I'm quite familiar with conda infrastructure, I'll try to implement it in a less fragile way (but thanks for the idea!).

PR #29, opened 2019-01-29
Rewrite build system, fix hbool_t, remove libhdf5-lib crate

### LINKED ISSUE #28: hbool_t on Windows and accessing HDF5 config header
So, we currently have `hbool_t` mapped as `c_uint` which turns out to be wrong because on Linux / OSX it's an alias of C `bool` which is 1-byte. As a result, some struct layouts get completely jumbled (which is pretty hard to check other than observing stuff failing in obscure ways).

Once I've changed it to 1 byte, it worked out perfectly on Unix, but some Windows tests started failing on AppVeyor (thanks god we have some kind of an actual test suite...). Took me a while to narrow down the issue (I had to install Windows with MSVS in a virtual machine...), turns out to be this:

```c
#ifdef H5_HAVE_STDBOOL_H
  #include <stdbool.h>
#else /* H5_HAVE_STDBOOL_H */
  #ifndef __cplusplus
    #if defined(H5_SIZEOF_BOOL) && (H5_SIZEOF_BOOL != 0)
      #define bool    _Bool
    #else
      #define bool    unsigned int
    #endif
    #define true    1
    #define false   0
  #endif /* __cplusplus */
#endif /* H5_HAVE_STDBOOL_H */
typedef bool hbool_t;
```

Now, the way it works with HDF5 builds, all those `H5_*` constants are auto-generated by the CMake process at build time, and dumped into a special header, `H5pubconf.h`, which is distributed along with the binaries and the source headers. It appears that whichever machine they were building this on, `stdbool.h` header was not accessible (which is somewhat weird because it has been available in MSVS since v2013). Apparently the CMake process has been [fixed upstream](https://github.com/live-clones/hdf5/commit/6f6d0921c0592c202bae15dab1b05c921aef0043) about two weeks ago, so it will now use assume C99, and thus use proper bool type – however, we can't make use of it since we have to support all the older versions of the library.

We could just say, of course,

```rust
#[cfg(target_env = "msvc")]
pub type hbool_t = c_uint;
#[cfg(not(target_env = "msvc"))]
pub type hbool_t = u8;
```

... but this is generally wrong, dodgy and fragile. Because, for instance, 3 years ago it looked like this:

```c
typedef unsigned int hbool_t;
```

Unfortunately, all those `H5_*` constants are not linked in into the library itself, so we can't access them the same way we extract library version.

... which leads to another point -- having those constants available as compile-time things, or at least some of them, like `H5_HAVE_PARALLEL`, `H5_THREADSAFE`, `H5_HAVE_DIRECT` would allow us to provide proper conditionally-compiled bindings for those cases (MPI support, thread-safe build and direct driver support, respectively).

Summarizing:

- Pros:
  - we gain access to auto-generated information about how the library was built
  - we can provide conditional bindings for MPI stuff, direct driver support, thread-safe builds etc
  - types like `hbool_t` will be properly bound
- Cons:
  - we have to assume `include/H5pubconf.h` always exists, we won't be able to build the crate without it (unless we make some assumptions which I'd rather not make at build time). This seems reasonable because the library is always distributed with the headers. Which means we'll have to search for the header kind of the same way we search for the library files; which might mean hard-coding library directory structure layout on windows and linux/osx.
  - we will either have to compile a C program into a crate (e.g. via using cc-rs; this implies access to a C compiler), or parse the header ourselves manually.

Options available:

1. Compile a simple C program that dumps all this stuff (or provides it as a crate) via [`cc-rs`](https://github.com/alexcrichton/cc-rs). Now this assumes existence of a C compiler on the current platform, I'm not sure but that sounds like an excessively harsh requirement.
2. Parse this C header manually with regex which shouldn't be too bad in theory  since it's just a bunch of `#define`s and not much else). We could probably even auto-generate a Rust module out of it with all those constants as `pub const` variables. We would obviously only be able to parse definitions not containing any expressions (that is, literals only).

aldanor:
- The build script has been completely rewritten; below are the major points.
- The env vars the build script reacts to are now `HDF5_DIR` and `HDF5_VERSION`.
- The build script now locates and parses `H5pubconf.h` (autogenerated header) via `bindgen`, extracting the library version and several preprocessor definitions out of it to be used later.
- This allows us to fix #28 (`hbool_t` is now defined conditionally).
- `h5_have_parallel`, `h5_have_threadsafe` and `h5_have_direct` are exposed for downstream use.
- The build script now loads the HDF5 library via `libloading` and verifies its version (we would probably still have to do this even if linking is static).
- `libhdf5-lib` crate is removed entirely, its contents are merged into `libhdf5-sys`.
- `pkg-config` now only runs on Linux.
- Added a few known Linux locations (when apt-installed on Ubuntu).
- Run `brew --prefix` on macOS; this supports both @1.8 and @1.10 bottles.
- On Windows, search the registry (via `winreg`) for official installations. 
- Dynamic linking vs conda hdf5 now works on all platforms and tested on CI.

Expand the CI test suite greatly (in particular, link against conda on all platforms along with other methods like apt, brew and official installers; build both HDF5 1.8 and HDF5 1.10 on all platforms):

| Rust | Linux  | macOS | Windows (MSVC) |
| - | - | - | - |
| stable | 1.10.4 (conda/xenial) | 1.8.21 (brew) | 1.8.21 (msi) |
| beta | 1.8.16 (apt/xenial) | 1.10.4 (brew) | 1.10.0 (msi) |
| nightly  | 1.8.11 (apt/trusty) | 1.8.9 (conda) | 1.10.4 (conda) |

### DIFF
--- a/libhdf5-sys/build.rs
+++ b/libhdf5-sys/build.rs
@@ -113,7 +113,7 @@ fn validate_runtime_version(config: &Config) {
     let mut link_paths = config.link_paths.clone();
     if cfg!(all(unix, not(target_os = "macos"))) {
         if let Some(ldv) = run_command("ld", &["--verbose"]) {
-            let re = Regex::new(r#"SEARCH_DIR\("=(?P<path>[^"]+)"\)"#).unwrap();
+            let re = Regex::new(r#"SEARCH_DIR\("=?(?P<path>[^"]+)"\)"#).unwrap();
             println!("Adding extra link paths (ld)...");
             for caps in re.captures_iter(&ldv) {
                 let path = &caps["path"];


aldanor:
@pmarks @Enet4 if you have any chance to test this out - would be much appreciated; if I don't hear from anyone I'll merge it in since it seems like it works on all platforms it's being tested on

Enet4:
I just ran `cargo test` real quick on this branch, I'm afraid it did not build successfully:

```
error: failed to run custom build command for `libhdf5-sys v0.2.0 (/home/enet4/hdf5-rs/libhdf5-sys)`
process didn't exit successfully: `/home/enet4/hdf5-rs/target/debug/build/libhdf5-sys-6be31ede1e344f33/build-script-build` (exit code: 101)
--- stdout
Attempting to find HDF5 via pkg-config...
Found HDF5 headers at:
    "/usr/include"
Adding to link path:
    "/usr/lib/x86_64-linux-gnu"
Parsing HDF5 config from:
    "/usr/include/H5pubconf.h"

--- stderr
thread 'main' panicked at 'Unable to infer HDF5 library runtime version.', libhdf5-sys/build.rs:144:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

I'm on Manjaro Linux (Arch-based), with HDF5 installed via pacman:

```sh
pacman -Si hdf5
```
```
Repository      : community
Name            : hdf5
Version         : 1.10.4-1
Description     : General purpose library and file format for storing scientific data
Architecture    : x86_64
URL             : https://www.hdfgroup.org/hdf5
Licenses        : custom
Groups          : None
Provides        : hdf5-cpp-fortran
Depends On      : zlib  libaec  bash
Optional Deps   : None
Conflicts With  : None
Replaces        : hdf5-cpp-fortran
Download Size   : 3,48 MiB
Installed Size  : 19,11 MiB
Packager        : Bruno Pagani <archange@archlinux.org>
Build Date      : seg 22 out 2018 12:03:08 WEST
Validated By    : MD5 Sum  SHA-256 Sum  Signature
```

aldanor:
@Enet4 thanks for testing!

So, not everything is lost, the following did work:
- pkg-config was called and its info parsed
- headers were located 
- H5pubconf.h header was parsed successfully
- library path was figured out

What didn’t work: loading the library .so via libloading to figure out its version. It’s hard to tell why it failed without debugging locally, options are:
- libhdf5.so wasn’t there at the link path (could you check? Is there anything arch/pacman special?)
- libloading couldn’t open the .so for some reason
- the .so was opened but H5get_libversion() failed (highly unlikely)

Note: if you run `cargo build/test -vv`, you can observe all of the output of the build script.

aldanor:
By the way, I think we can try and replace one of the xenial/apt Travis environments with Arch/pacman - I’ll take a look a bit later.

Enet4:
All right, another set of quick commands:

`whereis libhdf5`:
```
libhdf5: /usr/lib/libhdf5.a /usr/lib/libhdf5.settings /usr/lib/libhdf5.so
```

`cargo test -vv`

```
[libhdf5-sys 0.2.0] Attempting to find HDF5 via pkg-config...
[libhdf5-sys 0.2.0] Found HDF5 headers at:
[libhdf5-sys 0.2.0]     "/usr/include"
[libhdf5-sys 0.2.0] Adding to link path:
[libhdf5-sys 0.2.0]     "/usr/lib/x86_64-linux-gnu"
[libhdf5-sys 0.2.0] Parsing HDF5 config from:
[libhdf5-sys 0.2.0]     "/usr/include/H5pubconf.h"
[libhdf5-sys 0.2.0] thread 'main' panicked at 'Unable to infer HDF5 library runtime version.', libhdf5-sys/build.rs:144:5
[libhdf5-sys 0.2.0] note: Run with `RUST_BACKTRACE=1` for a backtrace.
error: failed to run custom build command for `libhdf5-sys v0.2.0 (/home/enet4/hdf5-rs/libhdf5-sys)`
...
```

I'm afraid that I am not in a good position to go deep into this right now, I'm on an important deadline until Thursday.

aldanor:
No problems, thanks for helping.

I see the problem, ok - `libhdf5.so` is in `/usr/lib`, whereas link path apparently reported by pkgconfig is `/usr/lib/x86_64-linux-gnu`. The build script tries to find .so in the latter and obviously fails.

The question is, how can it figure out about `/usr/lib` in a clean way? Hmm....

pmarks:
@aldanor very cool!  I tested it against my current macOS install (brew), and followed the conda directions in our linux environment. Both worked smoothly!   This is great for our near-term use cases.

aldanor:
@Enet4 the build script now looks for `ld --verbose` path, could you give it a try?

Enet4:
I'm afraid that it did not quite do the trick yet:

```
[libhdf5-sys 0.2.0] Attempting to find HDF5 via pkg-config...
[libhdf5-sys 0.2.0] Found HDF5 headers at:
[libhdf5-sys 0.2.0]     "/usr/include"
[libhdf5-sys 0.2.0] Adding to link path:
[libhdf5-sys 0.2.0]     "/usr/lib/x86_64-linux-gnu"
[libhdf5-sys 0.2.0] Parsing HDF5 config from:
[libhdf5-sys 0.2.0]     "/usr/include/H5pubconf.h"
[libhdf5-sys 0.2.0] Looking for HDF5 library binary...
[libhdf5-sys 0.2.0] Adding extra link paths (ld)...
[libhdf5-sys 0.2.0] thread 'main' panicked at 'Unable to infer HDF5 library runtime version (can't find the binary).', libhdf5-sys/build.rs:158:5
[libhdf5-sys 0.2.0] note: Run with `RUST_BACKTRACE=1` for a backtrace.
error: failed to run custom build command for `libhdf5-sys v0.2.0 (/home/enet4/hdf5-rs/libhdf5-sys)`
process didn't exit successfully: `/home/enet4/hdf5-rs/target/debug/build/libhdf5-sys-fc1258915cb468e0/build-script-build` (exit code: 101)
--- stdout
Attempting to find HDF5 via pkg-config...
Found HDF5 headers at:
    "/usr/include"
Adding to link path:
    "/usr/lib/x86_64-linux-gnu"
Parsing HDF5 config from:
    "/usr/include/H5pubconf.h"
Looking for HDF5 library binary...
Adding extra link paths (ld)...

--- stderr
thread 'main' panicked at 'Unable to infer HDF5 library runtime version (can't find the binary).', libhdf5-sys/build.rs:158:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

At this point it might be worth understanding why the `hdf5` package is adding a seemingly erroneous pk-config link path. I'll dump a few more outputs in case it helps.

`pkg-config --list-all | rg hdf`:

```
pkg-config --list-all | rg hdf
hdf5_hl_cpp-1.10.4             hdf5_hl_cpp - HDF5 (Hierarchical Data Format 5) Software Library
hdf5-1.10.4                    hdf5 - HDF5 (Hierarchical Data Format 5) Software Library
hdf5_hl-1.10.4                 hdf5_hl - HDF5 (Hierarchical Data Format 5) Software Library
hdf5_cpp-1.10.4                hdf5_cpp - HDF5 (Hierarchical Data Format 5) Software Library
```

`pkg-config --libs hdf5-1.10.4`

```
-lhdf5
```

It seems that this package configuration does not specify any link paths. Not that it should be a problem, considering that the shared objects are installed in the default system's lib directory.

aldanor:
The question is why no paths are added, seems like regex is not parsing anything successfully on your box from `ld —verbose`, extremely weird. You can check Travis logs for how it looks there - a bunch of paths like /usr/lib etc get added.

At this point I’d either need to install an arch virtual machine somewhere, or setup arch via docker on travis or something like that, either of which may take considerable time... so any help is greatly appreciated.

aldanor:
> Not that it should be a problem, considering that the shared objects are installed in the default system's lib directory.

Well, that’s exactly what we’re trying to do here with parsing `ld —verbose`, figure the system’s default lib directories...

aldanor:
@Enet4 wonder if you could upload the output of your local `ld --verbose` somewhere? (or try passing it through the regex in the build script yourself)

Enet4:
@aldanor I dumped the full output into a [gist](https://gist.github.com/Enet4/4d07803872e0dded31543d2c4b52ab29).

Enet4:
Yes, this one built successfully now! All tests (except the ones from hdf5-derive) also pass.

PR #35, opened 2019-03-07
0.5.0 release

### LINKED ISSUE #24: 0.5.0 release
So I've been thinking, although the crate is clearly in pre-alpha stage now, it has evolved enough for other folks to be able to contribute, plus it reached a somewhat usable state, with derive proc-macros done and basic dataset reading/loading working.

As such, I'm proposing to merge the current dev branch into master soon and release a 0.3.0a (where "a" underlines the crate's alpha state); motivation:

- People who visit github page wouldn't be thrown off by "last updated: 2.5 years ago" status (there's been over 350 new commits in the dev branches). Most of the traffic is from google, so it would help.
- Would hopefully also attract potential contributors (with master branch showing most recent progress).
- The current version would be cargo-installable without touching git
- hdf5-derive and hdf5-types crate names would be reserved on crates.io, currently they're not so anyone can take them.
- Update the docs (they are in messy state right now and most are missing, but at least the API would be sort-of-browsable).

Things to do:

- [x] Merge in the dataset slice read/write API (although it needs cleanup, more tests, refactoring and non-standard layout support - we'll refactor it later; the API isn't likely to change). (#20)
- [x] Merge of the latest work into master; clean up the branches.
- [x] Update the README with a simplest example involving derive macro and dataset reading / writing.
- [x] Update the changelog (going to be lengthy).
- [x] Fix compiletest on Travis so we can finally see some green builds (also they're currently failing on nightly due to https://github.com/rust-lang/rust/issues/57488; also there's an upstream bug in compiletest-rs with the latest nightly on OS X).
- [x] Fix the issue with `hbool_t` (#28); may require rewriting the current build system. (fixed in #29).
- [x] Make high-level objects cloneable, separate .copy() (#23) -- this is pretty simple to implement.
- [x] Finish the work on FCPL / FAPL, including various file driver backends and writing tests. (#31)
- [x] Resolve the story with potential crate renaming.

aldanor:
- The GitHub repository has been renamed to `aldanor/hdf5-rust` (need to double-check that travis/appveyor etc still work as intended)
- libhdf5-sys (libhdf5_sys) has been renamed to hdf5-sys (hdf5_sys)
- hdf5-rs (h5) has been renamed to hdf5
- Updated readmes, badges, etc
- Versions of all crates bumped to 0.3.0 (we initially planned to go with 0.3.0a but that would cause confusion since "0.3" is not <= than "0.3.0a", IIUC)
- Preparing all of this for release on crates.io

Closes #24.


aldanor:
The story with the existing crates is this: because of naming collision with previous hdf5/hdf5-sys crates and because of renaming, (1) all versions of the current hdf5/hdf5-sys will be yanked from crates.io; (2) all pre-0.3 versions of hdf5-rs, libhdf5-sys, libhdf5-lib, remutex, hdf5-types will be yanked from crates.io when we release v0.4.

This way, this version (0.3) will be the only earliest one remaining, and all versions of all hdf5-* crates on crates.io will be fully working and stable.

PR #40, opened 2019-04-10
Find the default HDF5 location in Fedora Linux

jneuff:
This solves issue 39.

### DIFF
--- a/hdf5-sys/build.rs
+++ b/hdf5-sys/build.rs
@@ -60,6 +60,7 @@ fn run_command(cmd: &str, args: &[&str]) -> Option<String> {
 #[allow(dead_code)]
 fn is_inc_dir<P: AsRef<Path>>(path: P) -> bool {
     path.as_ref().join("H5pubconf.h").is_file()
+    || path.as_ref().join("H5pubconf-64.h").is_file()
 }
 
 #[allow(dead_code)]
@@ -166,7 +167,8 @@ pub struct Header {
 impl Header {
     pub fn parse<P: AsRef<Path>>(inc_dir: P) -> Self {
         let inc_dir = inc_dir.as_ref();
-        let header = inc_dir.join("H5pubconf.h");
+        
+        let header = get_conf_header(inc_dir);
         println!("Parsing HDF5 config from:\n    {:?}", header);
 
         let contents = fs::read_to_string(header).unwrap();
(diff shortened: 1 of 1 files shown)


SuperFluffy:
This looks like fedora specific issue, not something that's related to how hdf5 itself is build or distributed.

I am not the maintainer, but I would personally not be in favor of merging distribution specific fixes.

jneuff:
> This looks like fedora specific issue, not something that's related to how hdf5 itself is built or distributed.
> 
> I am not the maintainer, but I would personally not be in favor of merging distribution specific fixes.

Generally, I agree with you. However, I am not sure about the reasons why HDF ist not picked up via pkg-conf. There seems to be a similar problem in Ubuntu, for which `hdf5-sys` already contains a fix. So the maintainer already went down that road, probably for a reason.

Let's hear some more opinions on this.

SuperFluffy:
Good to know, thanks.

aldanor:
Don't mind having Fedora-specific paths in the build script if it helps anyone. After all, the build script is one big hack (by purpose) and already contains Ubuntu-specific paths and macOS/Homebrew-specific stuff anyway, the point being it's expected to "just work" in most common setups which I believe it does (Fedora aside).

Two questions though:

- Why `H5pubconf-64.h`? First time I'm seeing this; is this a Fedora specific thing or an HDF5 one?
- Why does pkg-config method fail? Might be useful to know. Could you e.g. dump its output (from the build script) to check what it says?

// Ignore the nightly test failures; unrelated (to be fixed later)

jneuff:
>     * Why `H5pubconf-64.h`? First time I'm seeing this; is this a Fedora specific thing or an HDF5 one?
>

It seems to be a Fedora thing. Fedora 29 ships HDF 1.8.20, which seems a bit outdated. The header (H5pubconf-32/H5pubconf-64) depends on wether you install the x86 or the x64 version of the package. Presumably to make it possible to install both packages at the same time. However, I just tried to install the x86 version and it failed due to a conflicting `H5public.h` which both versions provide.

Maybe it would make sense to also extend this PR to accommodate the x86 version also. But I'm not sure how relevant that is.
 
>     * Why does pkg-config method fail? Might be useful to know. Could you e.g. dump its output (from the build script) to check what it says?
> 

Running `cargo build` on master gives me:

```
error: failed to run custom build command for `hdf5-sys v0.5.1 (/home/julius/scratch/hdf5-rs/hdf5-sys)`
process didn't exit successfully: `/home/julius/scratch/hdf5-rs/target/debug/build/hdf5-sys-167b9658896931f0/build-script-build` (exit code: 101)
--- stdout
Attempting to find HDF5 via pkg-config...

--- stderr
thread 'main' panicked at 'Unable to locate HDF5 root directory and/or headers.', hdf5-sys/build.rs:513:13
```

Running `pkg-config hdf5` manually just gives me nothing and `pkg-config --list-all` does not contain any related entries.

PR #43, opened 2019-05-28
allow the initial shape to be smaller than the chunk size, if the dataset is resizable

pmarks:
@aldanor This is allowed per the H5Pset_chunk docs, and works correctly for me. Seem reasonable?

### DIFF
--- a/src/hl/dataset.rs
+++ b/src/hl/dataset.rs
@@ -326,11 +326,14 @@ impl<T: H5Type> DatasetBuilder<T> {
                         "Invalid chunk: {:?} (all dimensions must be positive)",
                         dims
                     );
-                    ensure!(
-                        dims.iter().zip(shape.dims().iter()).all(|(&c, &s)| c <= s),
-                        "Invalid chunk: {:?} (must not exceed data shape in any dimension)",
-                        dims
-                    );
+
+                    if !self.resizable {
+                        ensure!(
+                            dims.iter().zip(shape.dims().iter()).all(|(&c, &s)| c <= s),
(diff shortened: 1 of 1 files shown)


SuperFluffy:
I am dealing with the same question right now. Can you point me to the relevant entries in the hdf5 docs?

As per https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetChunk:

> Chunk size cannot exceed the size of a fixed-size dataset. For example, a dataset consisting of a 5x4 fixed-size array cannot be defined with 10x10 chunks. 

I guess it's reasonable to assume that this kinda sorta implies that chunk sizes for extensible datasets *can* exceed the initial size. But did you find a doc that makes this explicit?

pmarks:
@SuperFluffy - I don't see docs that make it explicit.  I've relied on that behavior for years, and as you found there's an example that relies on it, so I'm pretty sure it's a supported case.

SuperFluffy:
I am using @pmarks' branch now, because I am aggregating my data during the run of my simulation, but don't want to store the data in an extra structure and then dump it at the end of a checkpoint. I also don't know the exact dimension of my dataset/number of entries, so allocating a monolithic slab also doesn't work.

I support merging this PR.

aldanor:
Fair enough - will try to merge it this weekend; thanks @pmarks

aldanor:
This has been merged but, interestingly enough, the tests for one of the Travis jobs froze on that particular test (resizable chunked dataset):

https://travis-ci.org/aldanor/hdf5-rust/jobs/558549732

Maybe it’s a coincidence though, I guess we’ll see.

Issue #14, opened 2017-07-06
Reading / writing


cathalgarvey:
This may be a stupid question, but I see the Dataset.read method being used in your example code in issues like #9 and I can't actually find documentation, or source code, for it. :)

I guess examples of use would be very helpful, as in #9, but I thought at least I could find the code to help me understand things! I have lots of questions, about what kinds of structured-types can be stored in datasets, whether datasets can be iterated to avoid loading them all into memory, etc. etc.; happy to answer these questions myself if I know where to look.

Thanks for what appears to be a very complete and powerful library!

Enet4:
I was once interested in this library for exactly that purpose. It happens that the code for reading from an existing dataset is in a separate branch, and is currently not documented. It might also be incomplete for certain data types.

Maybe the owner could use some help. We might be able to join forces around this functionality and document the working parts for future users.

aldanor:
@Enet4 @cathalgarvey Hi guys!

@Enet4 is correct, the type mapping is in a separate branch (types). I've rewritten it at least 3 times and it's pretty much complete -- supports all builtin types, tuple types, fixed-size array types, all 4 kinds of strings (fixed/varlen, ascii/utf-8 -- this already makes it support more atomics than e.g. pytables / h5py), plus it uses procedural macros to bind structs and nested structs, like so `#[derive(H5Type)]`. Last time I've had free time was this spring -- which is when I've managed to complete the type system.

This way, most of the super-low level stuff is actually done. The next one is Dataset API. I have a few implementations but I don't like neither of them too much for different reasons. One is `Dataset<T: H5Type>` where the element type is a type parameter. However, this way you lose the ability to inspect dataset header without caring for its type (e.g. check size / shape / ndim / filters and other stuff not depending on the type).  The other way is having two types for datasets, and a common trait, this way things get pretty complicated type-wise very fast. Another way is having `MaybeH5Type` trait and using `Dataset<T: MaybeH5Type>`, etc...

The ideal thing would be to get rid of all traits altogether (e.g. `Container`, `Location`) and use specializations instead; so you have things like `type Dataset = Object<DatasetID>`, `type UntypedDataset = Object<UntypedDatasetID>`. I've started doing work on that but then stumbled upon this unfortunate rustdoc bug https://github.com/rust-lang/rust/issues/32077 which made all docs disappear. Very sad since that would definitely be the best solution.

[...]

PR #49, opened 2019-07-17
File constructor / FileBuilder API rework

aldanor:
This is certainly a breaking change, but better sooner than later (initial discussion with @Enet4 a while ago in #14).

Main points:

- FAPL and FCPL bindings are pretty much 100% complete, it's a shame they couldn't have been used until now; this is now fixed, the new FileBuilder has full access to those (!).
- String-based modes like "r" and "w" are gone. Instead, you can just use `File::open()`, `File::open_rw()`, `File::create()`, `File::create_excl()` and `File::append()` (FileBuilder has those methods two as finishers).
- In case of need, there's also `File::open_as()` which mimics previous behaviour, but is strongly typed and doesn't use string-based modes either (FileBuilder has it too).
- Special cases like driver strings etc are gone from the new FileBuilder, instead you can access FAPL and FCPL builders from it in a few different ways (by importing an existing plist, by getting a mutable reference, or by providing a closure that does something).

Example 1:

```rust
// before
File::open("foo.h5", "r+")

// after
File::open_rw("foo.h5")
```

Example 2:

```rust
// before
File::with_options().mode("w").driver("core").filebacked(false).open(&filename)

// after
File::with_options().with_fapl(|p| p.core_filebacked(false)).create(&filename)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,38 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- `File::access_plist()` or `File::fapl()` to get file access plist.
+- `File::create_plist()` or `File::fcpl()` to get file creation plist.
+
+### Changed
+
+- Changed `File` constructors, getting rid of string access modes:
+  - `File::open(path, "r")` is now `File::open(path)`
+  - `File::open(path, "r+")` is now `File::open_rw(path)`
+  - `File::open(path, "w")` is now `File::create(path)`
+  - `File::open(path, "x" | "w-")` is now `File::create_exl(path)`
+  - `File::open(path, "a")` is now `File::append(path)`
+- Also added `File::open_as(path, mode)` which accepts the mode enum.
+- Rewritten `FileBuilder`: it no longer accepts userblock, driver etc;
+  all of these parameters can be set in the corresponding FAPL / FCPL:
+  - `FileBuilder::set_access_plist()` or `FileBuilder::set_fapl()` to
+    set the active file access plist to a given one.
+  - `FileBuilder::access_plist()` or `FileBuilder::fapl()` to get a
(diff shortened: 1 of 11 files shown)


SuperFluffy:
I wonder why you are not opting to represent the file constructor similar `std::fs::OpenOptions`?

aldanor:
@SuperFluffy 

Because it's simpler this way, I think? Not everything from there makes sense here (e.g. read is always enabled; append doesn't really have the same meaning etc; exclusive write is a new thing here) and as such blindly duplicating it doesn't make sense either.

Basically, my thought was - in `fs::OpenOptions` there's 6 options and 1 finalizer, `open()`. Multiple combinations of those options make sense for a file system, so there's up to 64 variants (some of which make sense, some don't, but what's important is that the number of valid options is certainly greater than 6.

Here, there's a few modes pre-defined in HDF5 and that's it ("append" synthetic mode is actually stolen from h5py and is just a convenient way of doing open-otherwise-create, it's not a built-in). Combining them doesn't make any sense whatsoever. So... instead of having 4/5 options and one finalizer, why not just have 5 finalizers (plus 1, the `open_as()`).

Plus, `File` gains the same finalizers, allowing you to do simply

```rust
File::open_rw("foo.h5")
```

without delving into docs of `FileBuilder` in order to figure out the chain of mutations you have to apply first. So, the user would only have to touch the builder if they want to change FAPL or FCPL - which is a much rarer case than just adjusting the file access mode.

aldanor:
Anyhow, while this may not be the final iteration (if anyone has strong opinions on how the API should look like), I'm strongly convinced that the proposed version is better than using `"r"` string to indicate reading and `"core"` string to indicated the core driver - which is what we have in the current master.

Here everything's strongly typed, and the most frequent use cases (open a file, create a new one, etc) become subjectively more simple and elegant.

PR #53, opened 2019-07-21
Added functionality to get all the members in a group

jbabyhacker:
If you merge this, could you please cut a new release?

### DIFF
--- a/hdf5-sys/src/h5o.rs
+++ b/hdf5-sys/src/h5o.rs
@@ -44,6 +44,20 @@ pub const H5O_HDR_ALL_FLAGS: c_uint = H5O_HDR_CHUNK0_SIZE
 pub const H5O_SHMESG_MAX_NINDEXES: c_uint = 8;
 pub const H5O_SHMESG_MAX_LIST_SIZE: c_uint = 5000;
 
+#[cfg(hdf5_1_10_3)]
+pub const H5O_INFO_BASIC: c_uint = 0x0001;
+#[cfg(hdf5_1_10_3)]
+pub const H5O_INFO_TIME: c_uint = 0x0002;
+#[cfg(hdf5_1_10_3)]
+pub const H5O_INFO_NUM_ATTRS: c_uint = 0x0004;
+#[cfg(hdf5_1_10_3)]
+pub const H5O_INFO_HDR: c_uint = 0x0008;
+#[cfg(hdf5_1_10_3)]
+pub const H5O_INFO_META_SIZE: c_uint = 0x0010;
+#[cfg(hdf5_1_10_3)]
+pub const H5O_INFO_ALL: c_uint =
+    H5O_INFO_BASIC | H5O_INFO_TIME | H5O_INFO_NUM_ATTRS | H5O_INFO_HDR | H5O_INFO_META_SIZE;
+
 #[repr(C)]
 #[derive(Copy, Clone, PartialEq, PartialOrd, Debug)]
 pub enum H5O_type_t {
--- a/src/hl/group.rs
+++ b/src/hl/group.rs
@@ -2,9 +2,13 @@ use std::fmt::{self, Debug};
 use std::ops::Deref;
 
 use hdf5_sys::{
+    h5::{hsize_t, H5_index_t, H5_iter_order_t},
     h5d::H5Dopen2,
     h5g::{H5G_info_t, H5Gcreate2, H5Gget_info, H5Gopen2},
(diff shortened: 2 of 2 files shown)


src/hl/group.rs
```diff
@@ -167,6 +169,36 @@ impl Group {
...
+    /// Non-recursively gets all the members in the group
+    pub fn get_member_names(&self) -> Result<Vec<String>> {
+        extern "C" fn get_members_callback(_id: hid_t,
+                                           name: *const c_char,
+                                           _info: *const H5L_info_t,
+                                           op_data: *mut c_void) -> herr_t {
+            let other_data: &mut Vec<String> = unsafe { &mut *(op_data as *mut Vec<String>) };
+
+            let c_str: &CStr = unsafe { CStr::from_ptr(name) };
+            let str_slice: &str = c_str.to_str().unwrap();
```
aldanor: There's `string_from_cstr()` which should be already imported in the prelude which woudl make this simpler

src/hl/group.rs
```diff
@@ -167,6 +169,36 @@ impl Group {
         let name = to_cstring(name)?;
         Dataset::from_id(h5try!(H5Dopen2(self.id(), name.as_ptr(), H5P_DEFAULT)))
     }
+
+    /// Non-recursively gets all the members in the group
```
aldanor: Maybe "Returns names of all the members in the group, non-recursively."? (also a missing period)

src/hl/group.rs
```diff
@@ -167,6 +169,36 @@ impl Group {
...
+    pub fn get_member_names(&self) -> Result<Vec<String>> {
+        extern "C" fn get_members_callback(_id: hid_t,
+                                           name: *const c_char,
+                                           _info: *const H5L_info_t,
```
aldanor: Could you please run `cargo fmt` so as to adhere to formatting in the crate?

src/hl/group.rs
```diff
@@ -167,6 +169,36 @@ impl Group {
...
+            let other_data: &mut Vec<String> = unsafe { &mut *(op_data as *mut Vec<String>) };
+
+            let c_str: &CStr = unsafe { CStr::from_ptr(name) };
+            let str_slice: &str = c_str.to_str().unwrap();
+            other_data.push(str_slice.to_string());
+
+            0 // Continue iteration
+        }
+
+        let callback_fn: H5L_iterate_t = Option::Some(get_members_callback);
```
aldanor: `Option::` is redundant

hdf5-sys/src/h5o.rs
```diff
@@ -44,6 +44,17 @@ pub const H5O_HDR_ALL_FLAGS: c_uint = H5O_HDR_CHUNK0_SIZE
 pub const H5O_SHMESG_MAX_NINDEXES: c_uint = 8;
 pub const H5O_SHMESG_MAX_LIST_SIZE: c_uint = 5000;
 
+pub const H5O_INFO_BASIC: c_uint = 0x0001;
```
aldanor: Have these constants remained the same throughout HDF5 versions? Just wanted to confirm that it would be the same from 1.8.5 to 1.10.5.

src/hl/group.rs
```diff
@@ -1,10 +1,12 @@
 use std::fmt::{self, Debug};
 use std::ops::Deref;
+use std::ffi::CStr;
```
aldanor: Can be dropped if you use the util helper (see comment below)

aldanor:
Also, maybe `get_member_names` could be renamed to simply `member_names`? I tried to avoid `get_` prefixes wherever possible.

src/hl/group.rs
jbabyhacker: Thanks for the suggestion. Done.

src/hl/group.rs
jbabyhacker: Done.

hdf5-sys/src/h5o.rs
jbabyhacker: I went through the code, verified these were unchanged an introduced in version 1.10.3. Added cfg attribute.

aldanor: Thanks! I must have missed it when updating.

aldanor:
Nice, thanks!

alexkirsz:
@aldanor Would it be possible to get a new release with this feature? The method is currently listed in the doc, but the last version on crates.io does not have it.

PR #54, opened 2019-08-01
Query threadsafe

jbabyhacker:
Exported call to determine if runtime version of the HDF5 library was built with --enable-threadsafe


jbabyhacker:
Tested by building hdf5 library from source with --enable-threadsafe specified, then running my changes and verified the result returned true. Ran the same code against a hdf5 library that wasn't built with threadsafe and the call to hdf5_threadsafe() returned false.

aldanor:
@jbabyhacker I'll add this in a branch myself, no problem.

(Travis failures seem unrelated btw)

PR #56, opened 2019-08-11
Added attribute support

jbabyhacker:


jbabyhacker:
@aldanor  Could you fix Travis CI issues?

aldanor:
@jbabyhacker Thanks for contribution - this will be a good start (although I'm currently in the process of completely rewriting the way datasets are handled, including dataset creation plists, filters support, external filters, dataspaces handling and all that - some of which will apply to attributes as well).

One important thing to note here: attributes can be attached to any named locations and not just datasets, they are not features of datasets.

(Also, could you please run cargo fmt, so that imports get sorted automatically, etc)

I will try looking into travis failures in the meanwhile.

jbabyhacker:
@aldanor Let me know if you need help with the rewrite. I browsed around the branches, but didn't see anything related to the rewrite. 

I am aware attributes can be attached to groups, I just didn't add that support in. 

Cargo fmt can be a little picky sometimes, it seem like I have to group imports for it to sort it properly. I'll double check in the future. 

I looked into fixing the Travis CI issue, but its an issue with Conda on Mac OS. If it was an issue on Linux, I could fix it easily.

aldanor:
> @aldanor Let me know if you need help with the rewrite. I browsed around the branches, but didn't see anything related to the rewrite.

It turned out to be a much bigger rework than I've initially planned (which was just DCPL which is done now) - I've added support for Blosc / LZF compression etc which still needs to be tested, rewrote extents/dataspace stuff etc. I'll try to push something compilable this weekend in a branch, but there's still quite a bit to be done. 

> I am aware attributes can be attached to groups, I just didn't add that support in.

I think you can just move the attribute-related methods from `Dataset` to `Location` and that will automatically "add support" for attributes in all location entities (including files, groups and datasets). That's the whole point of having deref-based object class hierarchy here.

> I looked into fixing the Travis CI issue, but its an issue with Conda on Mac OS. If it was an issue on Linux, I could fix it easily.

Yea, I'm on a mac right now, will try to check it out. I think they've reshuffled install locations in the hdf5 package, so the build.rs script may need to be altered too so it can pick it up.

jbabyhacker:
@aldanor Since the entire library is being refactored, I'm not going to update this PR. I'd rather just contribute to the new library. So let me know when the branch is pushed, and I'll add some tests to ensure everything is working correctly.

PR #59, opened 2019-09-28
Modify some errors in example of `README`

Axect:


Enet4:
It is not really an error, but the use of the latest API. See for example [`File::create`](https://github.com/aldanor/hdf5-rust/blob/master/src/hl/file.rs#L86). These changes are only downgrading to the old API.

aldanor:
@Enet4 is correct, thanks.

Apologies for the delay in version update (but it will be a big one).

PR #67, opened 2020-01-18
static build of hdf5

magnusuMET:
I am currently implementing a static build of `rust-netcdf`, and I need to obtain a static build of `hdf5`. This is strongly overlapping with the work in this crate, so I am wondering whether you'll accept these build changes.

The library user can now include `hdf-sys`, and get the include directory and library names through `DEP_HDF5_{KEY}` when needed for use in e.g. building `netCDF`.


gauteh:
Getting the following error when I try `cargo build --all-features`:

```
   Compiling mpi-sys v0.1.2
error: failed to run custom build command for `mpi-sys v0.1.2`

Caused by:
  process didn't exit successfully: `/home/gauteh/dev/hdf5-rust/target/debug/build/mpi-sys-5ac0430af6ce04a3/build-script-build` (exit code: 101)
--- stdout
TARGET = Some("x86_64-unknown-linux-gnu")
OPT_LEVEL = Some("0")
TARGET = Some("x86_64-unknown-linux-gnu")
HOST = Some("x86_64-unknown-linux-gnu")
TARGET = Some("x86_64-unknown-linux-gnu")
TARGET = Some("x86_64-unknown-linux-gnu")
HOST = Some("x86_64-unknown-linux-gnu")
CC_x86_64-unknown-linux-gnu = None
CC_x86_64_unknown_linux_gnu = None
HOST_CC = None
CC = Some("mpicc")
TARGET = Some("x86_64-unknown-linux-gnu")
HOST = Some("x86_64-unknown-linux-gnu")
CFLAGS_x86_64-unknown-linux-gnu = None
CFLAGS_x86_64_unknown_linux_gnu = None
HOST_CFLAGS = None
CFLAGS = None
DEBUG = Some("true")
running: "mpicc" "mpicc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-m64" "-Wall" "-Wextra" "-o" "/home/gauteh/dev/hdf5-rust/target/debug/build/mpi-sys-7bf89a262cc975b6/out/src/rsmpi.o" "-c" "src/rsmpi.c"
exit code: 0
TARGET = Some("x86_64-unknown-linux-gnu")
TARGET = Some("x86_64-unknown-linux-gnu")
HOST = Some("x86_64-unknown-linux-gnu")
AR_x86_64-unknown-linux-gnu = None
AR_x86_64_unknown_linux_gnu = None
HOST_AR = None
AR = None
TARGET = Some("x86_64-unknown-linux-gnu")
TARGET = Some("x86_64-unknown-linux-gnu")
running: "ar" "ar" "crs" "/home/gauteh/dev/hdf5-rust/target/debug/build/mpi-sys-7bf89a262cc975b6/out/librsmpi.a" "/home/gauteh/dev/hdf5-rust/target/debug/build/mpi-sys-7bf89a262cc975b6/out/src/rsmpi.o"
exit code: 0
TARGET = Some("x86_64-unknown-linux-gnu")
cargo:rustc-link-lib=static=rsmpi
cargo:rustc-link-search=native=/home/gauteh/dev/hdf5-rust/target/debug/build/mpi-sys-7bf89a262cc975b6/out
cargo:rustc-link-search=native=/usr//lib
cargo:rustc-link-search=native=/usr/lib/x86_64-linux-gnu/openmpi/lib
cargo:rustc-link-lib=mpi

--- stderr
warning: argument unused during compilation: '-L/usr//lib' [-Wunused-command-line-argument]
warning: argument unused during compilation: '-L/usr/lib/x86_64-linux-gnu/openmpi/lib' [-Wunused-command-line-argument]
/usr/lib/x86_64-linux-gnu/openmpi/include/mpi.h:229:10: fatal error: 'stddef.h' file not found
warning: argument unused during compilation: '-L/usr//lib' [-Wunused-command-line-argument], err: false
warning: argument unused during compilation: '-L/usr/lib/x86_64-linux-gnu/openmpi/lib' [-Wunused-command-line-argument], err: false
/usr/lib/x86_64-linux-gnu/openmpi/include/mpi.h:229:10: fatal error: 'stddef.h' file not found, err: true
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()', src/libcore/result.rs:1188:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

magnusuMET:
This PR only builds the `hdf5` c library, and exposes this through the `hdf5-sys`. If a static build of `hdf5` is needed, this needs to be added to the top `Cargo.toml`.

The PR also uses the `links` flags, which allows `hdf5-sys` to be removed as a build-time dependency, and prevents multiple versions of the same library to be linked at the same time.

gauteh:
Hi, I am getting a compilation error when enabling both `static` and `zlib`. The reason for trying that was that hdf5 was unable to find the deflate plugin (no plugin directory).

magnusuMET:
@gauteh Please try again!

gauteh:
Thanks, builds alright! Need to figure out #72 to get my code to build again, so not able to test whether it actually works right now. Also got a version conflict with rust-netcdf/vendor, but I guess that's just not updated yet either.

gauteh:
Alas, I spoke too soon. With both `static` and `zlib` enabled I get the following error. Maybe I am missing dependencies?

[hdf-build.log](https://github.com/aldanor/hdf5-rust/files/4219704/hdf-build.log)

magnusuMET:
The version conflict is not surprising, but can be circumvented by using the `patch` section. You can use the following in your toplevel `Cargo.toml` to override the dependency
```toml
[patch]
[patch.crates-io]
hdf5-sys = { git = "https://github.com/magnusuMET/hdf5-rust", branch = "feature/static_build" }
```

gauteh:
Thanks, looks like it works now!

magnusuMET:
@aldanor What is required before you can consider merging this PR?

gauteh:
Hi, running cargo build works fine, but trying to link causes an error on Mac OS X. This is from running `cargo test --features hdf5-sys/static,hdf5-sys/zlib`.

Note that just running `cargo test` works fine, and it seems like there are some compressed files as well in the test suite.

[hdf-build.log](https://github.com/aldanor/hdf5-rust/files/4221306/hdf-build.log)

magnusuMET:
@gauteh Thanks for the minimum repro case. I forgot to emit the link option, which does not seem to follow automatically from including `libz-sys`

gauteh:
Thanks, now everything is working (on linux at least).

hdf5-sys/build.rs
```diff
@@ -584,27 +590,117 @@ impl Config {
...
+    if feature!("PARALLEL").is_ok() {
+        // TODO: Find which dependencies to link
+        hdf5config.define("HDF5_ENABLE_PARALLEL", "ON");
+    }
+
+    let hdf5 = hdf5config
+        .profile("RelWithDebInfo") // TODO: detect opt-level and fix names of libraries
+        .build();
+
+    println!("cargo:rustc-link-search=native={}/lib", hdf5.display());
```
timfish: I did a bit of testing on Windows and I think it might be failing due to these hard-coded forwards slash paths

hdf5-sys/build.rs
```diff
@@ -584,27 +590,117 @@ impl Config {
...
+    println!("cargo:root={}", hdf5.display());
+
+    let hdf5_incdir = format!("{}/include", hdf5.display());
```
timfish: Another one here. 

Failing on Windows 10 with:
`error: could not find native static library `hdf5`, perhaps an -L flag is missing?`

timfish:
Actually, scrap my previous comments. The Windows static build fails because CMake is outputting `libhdf5.lib` and the Rust toolchain is expecting the library to be called `hdf5.lib` on Windows.

magnusuMET:
@timfish Thanks for investigating! I don't currently have access to a windows machine. Could you test using the latest commit?

hdf5-sys/build.rs
```diff
@@ -584,27 +590,123 @@ impl Config {
...
+        if feature!("HL").is_ok() || feature!("PARALLEL").is_ok() {
+            println!("cargo:warning=Unsupported selection of items. This is probably ok");
+            hdf5config.define("ALLOW_UNSUPPORTED", "ON");
+        }
+    }
+
+    if feature!("HL").is_ok() {
+        #[cfg(target_os = "windows")]
+        let hdf5_hl_lib = "libhdf5_hl";
+        #[cfg(not(target_os="Windows"))]
```
timfish: This should not be capitalised

hdf5-sys/build.rs
```diff
@@ -584,27 +590,123 @@ impl Config {
...
+
+    println!("cargo:rustc-link-search=native={}/lib", hdf5.display());
+    println!("cargo:root={}", hdf5.display());
+
+    let hdf5_incdir = format!("{}/include", hdf5.display());
+    println!("cargo:include={}", hdf5_incdir);
+
+    #[cfg(target_os = "windows")]
+    let hdf5_lib = "libhdf5";
+    #[cfg(not(target_os="Windows"))]
```
timfish: This should not be capitalised

magnusuMET:
@timfish Thanks for having a look! Force-pushed the fix on the `W`s

timfish:
That fixes it for Windows x64.

However, the build fails for `i686-pc-windows-msvc` with:
```
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_IEEE_F64LE_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17h8f0b035ad30a3492E
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_IEEE_F32LE_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17hf128fda671bab0eaE
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_STD_U64LE_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17h9c3d5d2a5e5b966bE
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5P_CLS_DATASET_CREATE_ID_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17haffd80dde87659daE
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_STD_I16LE_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17hdcb76de700608025E
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5P_CLS_FILE_CREATE_ID_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17ha89c752a9b8943aaE
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5P_CLS_FILE_ACCESS_ID_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17hfcf89fe0f8d6843bE
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_STD_I32LE_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17ha6aa53bbc633f91fE
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_NATIVE_INT_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17h6f725282d3fa1130E
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_C_S1_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17h1380b2202e3fa261E
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_STD_I64LE_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17h42fec7e26e560ad4E
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_NATIVE_INT8_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17h81a40de2b6c5ccbdE
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_STD_I8LE_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17h99d765115957db8eE
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_STD_U16LE_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17ha649864868ed11f7E
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_STD_U8LE_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17hd3c1e4b657380f68E
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5T_STD_U32LE_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17h93d15ab83c809e7eE
hdf5_writer.hdf5_writer.as269j1g-cgu.3.rcgu.o : error LNK2019: unresolved external symbol ___imp_H5P_CLS_LINK_CREATE_ID_g referenced in function __ZN3std4sync4once4Once9call_once28_$u7b$$u7b$closure$u7d$$u7d$17he066d73edb94ac6cE
          C:\Users\tim\Documents\Repositories\hdf5-writer\target\i686-pc-windows-msvc\release\deps\hdf5_writer.dll : fatal error LNK1120: 17 unresolved externals
```

Has the non-static build been tested working with 32bit Windows?

magnusuMET:
@timfish Is that build failure occuring on the non-static build? If that is the case, it should probably be in a separate issue

timfish:
@magnusuMET No, that failure is with the static build.

magnusuMET:
@timfish I'll get a windows computer and debug it sometime over the weekend

timfish:
It looks like the correct params are getting passed to CMake, ie. `-THost=x86` and `-AWin32`:

`[hdf5-sys 0.6.1] running: "cmake" "C:\\Users\\tim\\Documents\\Repositories\\hdf5-rust\\hdf5-sys\\source" "-G" "Visual Studio 16 2019" "-Thost=x86" "-AWin32" ....snip....`

magnusuMET:
Seems the bindings are expecting names on the form `__imp_NAME`, whilst the names should only have a single underscore when linking statically. When the compiler is following the name conventions for `x86`, an additional underscore is prepended, leading to the three underscores.

Where the `extern_static` macro is used, one must link to the expected name on static `x86` builds.

magnusuMET:
The `__imp_$NAME` was previously triggering on all builds for `msvc`, but should only be triggered for the dynamic `msvc` builds.

timfish:
@magnusuMET thanks for continuing to look into this, I owe you many beers!

Your most recent changes fix static build on `i686-pc-windows-msvc` but now `File::create` results in Access Violation on both x86 and x86_64 Windows. 

```
Exception thrown: read access violation.
***(*(__imp__ZN8hdf5_sys3h5p7globals19H5P_CLS_FILE_ACCESS17h7df273d68313cdafE))** was 0xFFFFFFFFFFFFFFFF. occurred

 	hdf5-test.exe!core::ops::function::FnOnce::call_once<fn() -> i64,()>(__int64(*)()) Line 232	Unknown
	hdf5-test.exe!lazy_static::lazy::{{impl}}::get::{{closure}}<i64,fn() -> i64>(lazy_static::lazy::{{impl}}::get::closure-0) Line 31	Unknown
 	hdf5-test.exe!std::sync::once::{{impl}}::call_once::{{closure}}<closure-0>(std::sync::once::{{impl}}::call_once::closure-0 *, bool) Line 264	Unknown
 	hdf5-test.exe!std::sync::once::Once::call_inner() Line 418	Unknown
 	hdf5-test.exe!std::sync::once::Once::call_once<closure-0>(lazy_static::lazy::{{impl}}::get::closure-0 self) Line 264	Unknown
 	[Inline Frame] hdf5-test.exe!hdf5::globals::{{impl}}::deref::__stability() Line 16	Unknown
 	hdf5-test.exe!hdf5::globals::{{impl}}::deref(hdf5::globals::H5P_FILE_ACCESS * self) Line 18	Unknown
 	hdf5-test.exe!hdf5::hl::plist::file_access::{{impl}}::try_new::{{closure}}(hdf5::hl::plist::file_access::{{impl}}::try_new::closure-0) Line 1477	Unknown
 	hdf5-test.exe!hdf5::sync::sync<core::result::Result<i64, hdf5::error::Error>,closure-0>(hdf5::hl::plist::file_access::{{impl}}::try_new::closure-0) Line 13	Unknown
 	hdf5-test.exe!hdf5::hl::plist::file_access::FileAccess::try_new() Line 1477	Unknown
 	hdf5-test.exe!hdf5::hl::plist::file_access::{{impl}}::finish::{{closure}}(hdf5::hl::plist::file_access::{{impl}}::finish::closure-0) Line 1467	Unknown
 	hdf5-test.exe!hdf5::sync::sync<core::result::Result<hdf5::hl::plist::file_access::FileAccess, hdf5::error::Error>,closure-0>(hdf5::hl::plist::file_access::{{impl}}::finish::closure-0 func) Line 13	Unknown
 	hdf5-test.exe!hdf5::hl::plist::file_access::FileAccessBuilder::finish() Line 1471	Unknown
 	hdf5-test.exe!hdf5::hl::file::{{impl}}::open_as::{{closure}}<str*>(hdf5::hl::file::{{impl}}::open_as::closure-1) Line 256	Unknown
 	hdf5-test.exe!hdf5::sync::sync<core::result::Result<hdf5::hl::file::File, hdf5::error::Error>,closure-1>(hdf5::hl::file::{{impl}}::open_as::closure-1 func) Line 13	Unknown
 	hdf5-test.exe!hdf5::hl::file::FileBuilder::open_as<str*>(str* self, hdf5::hl::file::OpenMode filename) Line 94	Unknown
 	hdf5-test.exe!hdf5::hl::file::File::open_as<str*>(str* filename, hdf5::hl::file::OpenMode mode) Line 102	Unknown
 	hdf5-test.exe!hdf5::hl::file::File::create<str*>(str* filename) Line 88	Unknown
 	hdf5-test.exe!hdf5_test::main() Line 11	Unknown
 	hdf5-test.exe!std::rt::lang_start::{{closure}}<core::result::Result<(), hdf5::error::Error>>(std::rt::lang_start::closure-0 *) Line 67	Unknown
 	[Inline Frame] hdf5-test.exe!std::rt::lang_start_internal::{{closure}}() Line 52	Unknown
 	hdf5-test.exe!std::panicking::try::do_call<closure-0,i32>() Line 292	Unknown
 	hdf5-test.exe!panic_unwind::__rust_maybe_catch_panic() Line 78	Unknown
 	[Inline Frame] hdf5-test.exe!std::panicking::try() Line 270	Unknown
 	[Inline Frame] hdf5-test.exe!std::panic::catch_unwind() Line 394	Unknown
 	hdf5-test.exe!std::rt::lang_start_internal() Line 51	Unknown
 	hdf5-test.exe!std::rt::lang_start<core::result::Result<(), hdf5::error::Error>>(core::result::Result<(), hdf5::error::Error>(*)() main, __int64 argc, unsigned char * * argv) Line 67	Unknown
 	hdf5-test.exe!main()	Unknown


```

magnusuMET:
@timfish The latest commit now checks if we are linking to a `dll` or not

timfish:
@magnusuMET that fixes the previous issue. 

I'm continuing with more testing and I've found another difference between architectures on Windows.

In my test binary I have the following code:
```rust
        let dataset = file
            .new_dataset::<f32>()
            .gzip(gz_level)
            .resizable(true)
            .chunk(chunk_size)
            .create(name, 0)
            .expect("Could not add dataset");
```
On `x86_64-pc-windows-msvc` this works and results in the expected file.
On `i686-pc-windows-msvc` I get the following panic:
```
thread 'main' panicked at 'Could not add dataset: Encoding is not enabled for filter: gzip'
```

magnusuMET:
@timfish Have you enabled the `zlib` feature of `hdf5-sys`?

timfish:
Yes, I've enabled `static` and `zlib`. It works fine on x64 but for some reason it says the encoding is missing at runtime on x86. The only this that differs between the two builds is the `--target`.

magnusuMET:
@timfish I have not been able to reproduce. To test I've added the above snippet as a new test binary, and ran
```sh
cargo +toolchain test --features "hdf5-sys/static,hdf5-sys/zlib"
```
where for toolchain I've used both windows x64 and x86. Both builds completed the gzip test (but x86 failed another unrelated test).

aldanor:
Apologies, this took me a while to review; I'll try to build it on osx tonight when I get to my home computer and comment more.

A few notes so far:

- Do I understand this correctly? The submodule is not pinned to a particular HDF5 commit, so if I build this today and a commit appears upstream, tomorrow my build is not going to be the same. Also, if a bug or an uncompileable commit is pushed upstream, my build is broken?
- The HDF5 version is not selectable? In which case it probably makes sense to pin it to 1.10.6 and note it in the docs.
- The entire HDF5 git source is stored with the crate? (it's pretty big...) An alternative would be to git clone/checkout manually I guess (which requires git to be installed).
- There's been some changes in master, could you please rebase the branch?
- Please don't do any version bumps here if possible, we'll bump versions of all crates separately when releasing.

magnusuMET:
Thanks @aldanor for having a look!

* The submodule is pinned fo commit `f5331844e142e4a456ea5452ce3a3980f8bbf57c`, which is the release of version 1.10.6. To update, one must manually upgrade the submodule and commit the new HEAD pointer to this repo.
* To have selectable versions one either needs to set up a patching system, or include parallel versions of the source directory. Is there any benefit to the user to downgrade the library version?
* The source code will be included on `crates.io`, which will inflate the crate size by about 19MB. We could split some of this into a separate  `src` crate which will only include the source and build script

magnusuMET:
Branch is now rebased with no version bumps

aldanor:
@magnusuMET 

> To have selectable versions one either needs to set up a patching system, or include parallel versions of the source directory. Is there any benefit to the user to downgrade the library version?

Well... then why 1.10.6 and not 1.12.0 which has been released a few days ago? I believe not everyone would want to jump on 1.12.0 straight away (e.g. given that there were issues with 1.10.0 that were fixed in later point releases). Also, 1.12 e.g. requires newer versions of cmake on all platforms.

> The source code will be included on crates.io, which will inflate the crate size by about 19MB. We could split some of this into a separate src crate which will only include the source and build script. We could look to the blas-src crate for inspiration on this.

I checked the blas-src family of crates (openblas-src etc), that looks like a reasonable approach, maybe we could do the same? E.g., have an `hdf5-src` crate with the source tree, so if you don't choose a static build, the crate is not inflated by 19MB. Would it be possible to maintain multiple source versions though? At least 1.10.* series and 1.12.* series perhaps, would that make sense? Would that inflate the whole 'static source' crate twofold?

hdf5-sys/build.rs
```diff
@@ -584,27 +593,123 @@ impl Config {
...
+        .define("HDF5_ENABLE_PARALLEL", "OFF")
+        .define("HDF5_ENABLE_DEPRECATED_SYMBOLS", "OFF")
+        .define("HDF5_ENABLE_THREADSAFE", "OFF")
+        .define("ALLOW_UNSUPPORTED", "OFF")
+        .define("HDF5_BUILD_HL_LIB", "OFF");
+
+    if feature!("ZLIB").is_ok() {
+        let zlib_include_dir = std::env::var_os("DEP_Z_INCLUDE").unwrap();
+        let mut zlib_header = std::env::split_paths(&zlib_include_dir).nth(0).unwrap();
+        zlib_header.push("zlib.h");
```
aldanor: Wonder what to do about zlib, if we're packaging 20mb of hdf5 source, shouldn't we just package zlib as well? I.e., when would the user choose to point at their own zlib.h?

hdf5-sys/build.rs
```diff
@@ -584,27 +593,123 @@ impl Config {
...
+
+    if feature!("THREADSAFE").is_ok() {
+        hdf5config.define("HDF5_ENABLE_THREADSAFE", "ON");
+        if feature!("HL").is_ok() || feature!("PARALLEL").is_ok() {
+            println!("cargo:warning=Unsupported selection of items. This is probably ok");
+            hdf5config.define("ALLOW_UNSUPPORTED", "ON");
+        }
+    }
+
+    if feature!("HL").is_ok() {
```
aldanor: Does hl make sense given that we don't provide bindings for it at this point?

hdf5-sys/build.rs
aldanor: Ah, apologies, I've misread this; does DEPS_Z_INCLUDE come from libz-sys/static? So it is self-sufficient then.

magnusuMET: Yeah, `DEP_{LIB}_{SOMETHING}` is included from the dependency. This way of passing flags up is enabled through `links` in `Cargo.toml`

hdf5-sys/build.rs
magnusuMET: Both `hl` and `deprecated` are included to reuse the library in the `netcdf` crate. These features are only needed to get the right build flags and linking. Since these are hidden behind feature flags on `hdf5-sys`, I would say they are obscured enough that a user would not expect them to be available.

magnusuMET:
> Why 1.10.6 and not 1.12.0 which has been released a few days ago? I believe not everyone would want to jump on 1.12.0 straight away (e.g. given that there were issues with 1.10.0 that were fixed in later point releases). Also, 1.12 e.g. requires newer versions of cmake on all platforms.

As far as I can tell, this crate has not been tested on the 1.12 branch, so I've selected the latest one that seems to be supported.

> I checked the blas-src family of crates (openblas-src etc), that looks like a reasonable approach, maybe we could do the same? E.g., have an `hdf5-src` crate with the source tree, so if you don't choose a static build, the crate is not inflated by 19MB. Would it be possible to maintain multiple source versions though? At least 1.10.* series and 1.12.* series perhaps, would that make sense? Would that inflate the whole 'static source' crate twofold?

One could create a crate for each series (`hdf5-src-110`, hdf5-src-112`), but I am afraid the selection of which version to use would be confusing. `blas` requires the end user to select the `blas` version to use.

Using the `hdf5-src` approach would also complicate passing the correct linking flags, I've haven't tried this approach yet.

gauteh:
Is it possible with cargo to have:

* hdf5-src released with 1.10.* and 1.12.*
* then depend on hdf5-src >= 1.10.* in hdf5-sys[static], which would select
1.12.* if possible?
* an end-user could then add a dependency themselves for hdf5-src<=1.12.*.
Would this not force hdf5-sys to use hdf5-src==1.10.* as well?

magnusuMET:
I am not sure whether that would work. To get the link flags from a crate, you would need to depend on `hdf5-src` in `hdf5-sys`, which requires pinning the version to publish on `crates.io` (I think we need to specify minor version?). We do not want to end up with two versions of `hdf5-sys`, as that requires two versions of `hdf5-src` etc.

I think the simplest option would be to only link to 1.10 for now, and set up more advanced options if requested from users.

magnusuMET:
I think this PR is feature complete now. The static feature is not yet enabled at the toplevel crate, but this can be enabled easily in the future.

magnusuMET:
@aldanor Any blockers for merging this PR?

gauteh:
Hi, I would also be very interested in this functionality! Please consider if ready for merging. Thanks everybody, for you work.

gauteh:
Hi, I just gave this a try with latest version and it seems like the source is not fetched automatically:

```
~/dev/hdf5-rust feature/static_build
❯ cargo build --features hdf5-sys/static
   Compiling hdf5-src v0.1.0 (/home/gauteh/dev/hdf5-rust/hdf5-src)
error: failed to run custom build command for `hdf5-src v0.1.0 (/home/gauteh/dev/hdf5-rust/hdf5-src)`

Caused by:
  process didn't exit successfully: `/home/gauteh/dev/hdf5-rust/target/debug/build/hdf5-src-4b06353fdccc7af5/build-script-build` (exit code: 101)
--- stdout
cargo:rerun-if-changed=build.rs
running: "cmake" "/home/gauteh/dev/hdf5-rust/hdf5-src/source" "-DHDF5_NO_PACKAGES=ON" "-DBUILD_SHARED_LIBS=OFF" "-DBUILD_TESTING=OFF" "-DHDF5_BUILD_TOOLS=OFF" "-DHDF5_BUILD_EXAMPLES=OFF" "-DHDF5_BUILD_JAVA=OFF" "-DHDF5_BUILD_FORTRAN=OFF" "-DHDF5_BUILD_CPP_LIB=OFF" "-DHDF5_ENABLE_PARALLEL=OFF" "-DHDF5_ENABLE_DEPRECATED_SYMBOLS=OFF" "-DHDF5_ENABLE_THREADSAFE=OFF" "-DALLOW_UNSUPPORTED=OFF" "-DHDF5_BUILD_HL_LIB=OFF" "-DCMAKE_INSTALL_PREFIX=/home/gauteh/dev/hdf5-rust/target/debug/build/hdf5-src-24cf632839058e3f/out" "-DCMAKE_C_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_C_COMPILER=/usr/bin/cc" "-DCMAKE_CXX_FLAGS= -ffunction-sections -fdata-sections -fPIC -m64" "-DCMAKE_CXX_COMPILER=/usr/bin/c++" "-DCMAKE_BUILD_TYPE=RelWithDebInfo"

--- stderr
CMake Error: The source directory "/home/gauteh/dev/hdf5-rust/hdf5-src/source" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
thread 'main' panicked at '
command did not execute successfully, got: exit code: 1

build script failed, must exit now', /home/gauteh/.cargo/registry/src/github.com-1ecc6299db9ec823/cmake-0.1.42/src/lib.rs:861:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

magnusuMET:
The source will be fetched with the `hdf5-src` crate when this is uploaded, so this will not be an issue for the end users. There should probably be a notice for developers somewhere that they need to pull all submodules when building statically from source, using for example
```sh
git submodule update --init --recursive
```

magnusuMET:
Build failure seems to be unrelated to this PR

gauteh:
Hi, I have been using this since January without problems. It makes working with the HDF5 library much smoother and more reliable, since the default hdf5 in ubuntu is missing some essential features for me and it is easier to predict which features are available.

timfish:
Same here. 

We've been using this in a production library on Windows, macOS and Linux. It's even building on Raspberry Pi once clang was installed.

aldanor:
@magnusuMET Thanks for working on this and apologies for being absent for away (with covid and all, been completely detached from real life for a while), I'll try to take a detailed look at this again on the weekend and perhaps try to merge it.

magnusuMET:
Thanks @aldanor! I see that we also have several other exciting PRs in the backlog. Would you be open to adding another contributor to help with the workload?

aldanor:
@magnusuMET I think the easiest way to speed things up (and that's the reason for not implementing attributes) is for people to help me with HUMONGOUS branch that I have 95%-done which

- covers entire DCPL plists
- adds support for virtual datasets
- rewrites entire extents/dimension to allow for arbitrary hdf5-supported extents, including custom unlimited shapes
- adds support for custom selections
- covers entire DAPL plists
- adds support for LZF filters and compression
- adds support for Blosc filters and compression
- rewrites Dataset and DatasetBuilder from scratch to support all of the above

It's currently in feature/dcpl but it's not the very latest version. I can work on pushing the latest buildable version into a PR and list a number of things that will have to be completed and/or tested.

magnusuMET:
@aldanor I would be very happy to help review such a PR!

hdf5-sys/build.rs
```diff
@@ -9,6 +9,12 @@ use std::str;
 
 use regex::Regex;
 
+macro_rules! feature {
+    ($feature:expr) => {
+        std::env::var(concat!("CARGO_FEATURE_", $feature))
+    };
+}
```
aldanor, line 16:
```suggestion
fn feature_enabled(feature: &str) -> bool {
    env::var(format!("CARGO_FEATURE_{}", feature)).is_ok()
}
```

hdf5-sys/build.rs
```diff
@@ -571,6 +577,16 @@ impl Config {
         }
         println!("cargo:rerun-if-env-changed=HDF5_DIR");
         println!("cargo:rerun-if-env-changed=HDF5_VERSION");
+        if cfg!(target_env = "msvc") {
```
aldanor, line 580:
```suggestion

        if cfg!(target_env = "msvc") {
```

hdf5-sys/build.rs
```diff
@@ -571,6 +577,16 @@ impl Config {
...
+            println!("cargo:is_dll=1");
+        }
+        println!("cargo:include={}", self.inc_dir.to_str().unwrap());
+
+        println!("cargo:library=hdf5");
+
+        if feature!("HL").is_ok() {
```
aldanor, line 587:
```suggestion
        if feature_enabled("HL") {
```

hdf5-sys/build.rs
```diff
@@ -580,27 +596,68 @@ impl Config {
...
     }
 }
 
 fn main() {
-    let mut searcher = LibrarySearcher::new_from_env();
-    searcher.try_locate_hdf5_library();
-    let config = searcher.finalize();
-    println!("{:#?}", config);
-    config.emit_link_flags();
+    if feature!("STATIC").is_ok() && !std::env::var_os("HDF5_DIR").is_some() {
```
aldanor, line 621:
```suggestion
    if feature_enabled("STATIC") && !env::var("HDF5_DIR").is_ok() {
```

hdf5-sys/build.rs
```diff
@@ -580,27 +596,68 @@ impl Config {
...
+fn get_build_and_emit() {
+    println!("cargo:rerun-if-changed=build.rs");
+
+    if feature!("ZLIB").is_ok() {
+        let zlib_lib = std::env::var("DEP_HDF5SRC_ZLIB").unwrap();
+        println!("cargo:zlib={}", &zlib_lib);
+        let zlib_lib_header = std::env::var("DEP_HDF5SRC_ZLIB").unwrap();
+        println!("cargo:zlib={}", &zlib_lib_header);
+        println!("cargo:rustc-link-lib=static={}", &zlib_lib);
+    }
```
aldanor, line 642:
```suggestion
    if feature_enabled("ZLIB") {
        let zlib_lib = env::var("DEP_HDF5SRC_ZLIB").unwrap();
        println!("cargo:zlib={}", &zlib_lib);
        let zlib_lib_header = env::var("DEP_HDF5SRC_ZLIB").unwrap();
        println!("cargo:zlib={}", &zlib_lib_header);
        println!("cargo:rustc-link-lib=static={}", &zlib_lib);
    }
```

hdf5-sys/build.rs
```diff
@@ -580,27 +596,68 @@ impl Config {
...
+
+    if feature!("HL").is_ok() {
+        let hdf5_hl_lib = std::env::var("DEP_HDF5SRC_HL_LIBRARY").unwrap();
+        println!("cargo:rustc-link-lib=static={}", &hdf5_hl_lib);
+        println!("cargo:hl_library={}", &hdf5_hl_lib);
+    }
```
aldanor, line 648:
```suggestion
    if feature_enabled("HL") {
        let hdf5_hl_lib = env::var("DEP_HDF5SRC_HL_LIBRARY").unwrap();
        println!("cargo:rustc-link-lib=static={}", &hdf5_hl_lib);
        println!("cargo:hl_library={}", &hdf5_hl_lib);
    }
```

hdf5-sys/build.rs
```diff
@@ -580,27 +596,68 @@ impl Config {
...
+
+    let hdf5 = std::env::var("DEP_HDF5SRC_ROOT").unwrap();
+    println!("cargo:root={}", &hdf5);
+    let hdf5_incdir = std::env::var("DEP_HDF5SRC_INCLUDE").unwrap();
+    println!("cargo:include={}", &hdf5_incdir);
+    let hdf5_lib = std::env::var("DEP_HDF5SRC_LIBRARY").unwrap();
+    println!("cargo:library={}", &hdf5_lib);
```
aldanor, line 655:
```suggestion
    let hdf5_root = env::var("DEP_HDF5SRC_ROOT").unwrap();
    println!("cargo:root={}", &hdf5_root);
    let hdf5_incdir = env::var("DEP_HDF5SRC_INCLUDE").unwrap();
    println!("cargo:include={}", &hdf5_incdir);
    let hdf5_lib = env::var("DEP_HDF5SRC_LIBRARY").unwrap();
    println!("cargo:library={}", &hdf5_lib);
```

hdf5-sys/build.rs
```diff
@@ -580,27 +596,68 @@ impl Config {
...
+
+    println!("cargo:rustc-link-search=native={}/lib", &hdf5);
```
aldanor, line 657:
```suggestion
    println!("cargo:rustc-link-search=native={}/lib", &hdf5_root);
```

@magnusuMET Is using "/" ok here? E.g. on Windows

Cargo.toml
```diff
@@ -39,6 +39,3 @@ rand = { version = "0.7", features = ["small_rng"] }
 regex = "1.3"
 scopeguard = "1.0"
 tempdir = "0.3"
-
-[build-dependencies]
-hdf5-sys = { path = "hdf5-sys", version = "0.6.1" }  # !V
```
aldanor, line 44: @magnusuMET So it's no longer a build dependency?

hdf5-src/build.rs
```diff
@@ -0,0 +1,80 @@
...
+        hdf5config.define("HDF5_BUILD_HL_LIB", "ON");
+    }
+
+    if feature!("PARALLEL").is_ok() {
+        // TODO: Find which dependencies to link
+        hdf5config.define("HDF5_ENABLE_PARALLEL", "ON");
+    }
+
+    let hdf5 = hdf5config
+        .profile("RelWithDebInfo") // TODO: detect opt-level and fix names of libraries
```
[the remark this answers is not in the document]
magnusuMET: The profile is automatically detected, which is overridden here, so we don't have to rebuild when switching between `debug` and `release`. The library name changes depending on what profile we compile with, `libhdf5-debug` or `libhdf5` (or something like that).

hdf5-src/build.rs
```diff
@@ -0,0 +1,80 @@
...
+        #[cfg(target_os = "windows")]
+        let hdf5_hl_lib = "libhdf5_hl";
+        #[cfg(not(target_os = "windows"))]
+        let hdf5_hl_lib = "hdf5_hl";
+        println!("cargo:hl_library={}", hdf5_hl_lib);
+        hdf5config.define("HDF5_BUILD_HL_LIB", "ON");
+    }
+
+    if feature!("PARALLEL").is_ok() {
+        // TODO: Find which dependencies to link
```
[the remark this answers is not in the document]
magnusuMET: We would now require an MPI aware compiler, which we could get from `rsmpi`, or through some environment flag.

.travis.yml
```diff
@@ -63,7 +63,7 @@ script:
   fi
 - cargo clean
 - cargo build -vv
-- cargo test -v --all
+- cargo test -v --all --exclude hdf5-src
```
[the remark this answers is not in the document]
magnusuMET: Travis does not pull submodules, which means this crate will not build. We could set up github actions for the static builds, that seems to be quite flexible.

hdf5-sys/build.rs
```diff
@@ -571,6 +577,16 @@ impl Config {
         }
         println!("cargo:rerun-if-env-changed=HDF5_DIR");
         println!("cargo:rerun-if-env-changed=HDF5_VERSION");
+        if cfg!(target_env = "msvc") {
+            println!("cargo:is_dll=1");
```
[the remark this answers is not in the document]
magnusuMET: This is only emitted when searching for the library, and not the build path exposed here. `is_win_dll` is probably the clearest, although I do not know which windows targets the special casing to enforce `dllimport` is required.

hdf5-sys/src/h5e.rs
```diff
@@ -84,7 +84,7 @@ extern "C" {
 
 pub use self::globals::*;
 
-#[cfg(not(target_env = "msvc"))]
+#[cfg(not(all(target_env = "msvc", not(feature = "static"))))]
```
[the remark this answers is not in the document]
magnusuMET: Maybe something along `dll_name_mangling`? Is this an upstream bug that `__imp_` is not used automatically? Is there a `dllimport` missing on extern statics?

hdf5-src/build.rs
```diff
@@ -0,0 +1,80 @@
...
+        let zlib_include_dir = std::env::var_os("DEP_Z_INCLUDE").unwrap();
+        let mut zlib_header = std::env::split_paths(&zlib_include_dir).nth(0).unwrap();
+        zlib_header.push("zlib.h");
+        let zlib_lib = "z";
+        hdf5config
+            .define("HDF5_ENABLE_Z_LIB_SUPPORT", "ON")
+            .define("H5_ZLIB_HEADER", &zlib_header)
+            .define("ZLIB_STATIC_LIBRARY", &zlib_lib);
+        println!("cargo:zlib_header={}", zlib_header.to_str().unwrap());
+        println!("cargo:zlib={}", zlib_lib);
```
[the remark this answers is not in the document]
magnusuMET: We really want to keep the `OsString`, as some paths are not valid `UTF-8`.

hdf5-sys/build.rs
magnusuMET: This is supported, although I can't seem to find the documentation for this.

Cargo.toml
magnusuMET: Yes, we only need to build this library once, not also during build phase. build scripts are built as debug, so this is really necessary when compiling also `hdf5-src`!

magnusuMET:
@aldanor Your code suggestions has been implemented! Some further questions:
* How much do we care about correctness in paths (utf-8)? This would be easy to fix on unix, but I don't know how this works on windows.
* Should we remove the `parallel` feature before it is requested?
* dll handling and naming + extra indirection required (`is_dll`)

I'll setup some github actions to test the static branch on all platforms. I'll do some squashing on the commit history when we've resolved some more of the current issues.

hdf5-src/build.rs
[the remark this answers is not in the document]
aldanor: I don't have much experience with it aside from creating bindings for mpi-enabled HDF5. It looks like in current state you won't be able to build static mpi-enabled HDF5 anyway, so there's two options - either fully work it out and test it (which is quite some work?), or drop the static parallel feature (i.e. raise cargo error if mpio and static are both enabled) for now, with a note that it can be implemented later if needed?

aldanor:
>     * How much do we care about correctness in paths (utf-8)? This would be easy to fix on unix, but I don't know how this works on windows.

Typically, it's ok for build scripts to throw on unwrapping OsStr to String conversion if that's what you mean? (there's also joining paths via "/" which I've commented on above, I don't know if that's ok or not on Windows in cargo)=

>     * Should we remove the `parallel` feature before it is requested?

See my comments above - we can leave a note that it's possible to implement it, and disable the combination of mpio and static?

>     * dll handling and naming + extra indirection required (`is_dll`)

What about it? (see a few of my questions above - a bit confusing re: what "is_dll" actually means)

> I'll setup some github actions to test the static branch on all platforms. I'll do some squashing on the commit history when we've resolved some more of the current issues.

Currently, it's Travis + AppVeyor... I was planning on moving the entire thing to GitHub Actions and ditching travis/appveyor but it would be outside of scope of this PR (help welcome of course).

magnusuMET:
During the CI I found and fixed some minor problems. The `hdf5-derive` crate can also not be tested when building statically, the `hdf5` versions end up incompatible (one built statically, another built dyn).

magnusuMET:
> Typically, it's ok for build scripts to throw on unwrapping OsStr to String conversion if that's what you mean? (there's also joining paths via "/" which I've commented on above, I don't know if that's ok or not on Windows in cargo)

The path separator is OK, the windows NT layer translates this to a backslash. Not all valid paths are valid UTF-8, but this implementation only supports the ones with valid UTF-8. On windows this is a bit tricky, since there is a conversion between codepages, and we would need to investigate what `hdf5-c` does under these circumstances. On unix we could use the bytes from `OsStr` (pad with a zero) as a valid filename. This could be a separate issue, and possibly unlikely to show up?

hdf5-src/build.rs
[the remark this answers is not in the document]
magnusuMET: Seems this postfix is added when passing CMAKE_BUILD_TYPE=Debug, which happens on opt-level=0

magnusuMET:
@aldanor I did some serious cleanup of the commit history, and added most of your suggestions. I think this PR is ready to go!

The results of the CI can be observed at https://github.com/magnusuMET/hdf5-rust/actions, this PR builds on linux, macos, and windows (msvc and gnu).

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Added
 
+- HDF5 C library can now be built and statically linked to `hdf5-sys`
+  crate by enabling `hdf5-sys/static` feature.
```
aldanor, line 8:
```suggestion
- HDF5 C library can now be built and statically linked to `hdf5-sys`
  crate by enabling `hdf5-sys/static` feature (cmake required).
```

README.md
```diff
@@ -91,6 +91,9 @@ a modern Rust compiler (e.g. of version 1.40 or later).
 Required HDF5 version is 1.8.4 or newer. The library doesn't have to be built with
 threadsafe option enabled.
 
+The HDF5 C library can also be built from source and linked in statically by 
+enabling `hdf5-sys/static` feature.
```
aldanor, line 95:
```suggestion
The HDF5 C library can also be built from source and linked in statically by 
enabling `hdf5-sys/static` feature (cmake required).
```

hdf5-sys/README.md
```diff
@@ -0,0 +1,32 @@
...
+but it may be incremented later.
+
+## Crate features
+
+Features propagated upwards will be detected based on features available for the chosen library. The library selected can always be overridden by specifying the environment variable `HDF5_DIR`.
+
+General features:
+
+- `mpio`: enable MPI support (in static mode, also enable it in the HDF5 C library)
+- `static`: build HDF5 C library from source (see below for a list of options).
```
aldanor, line 22:
```suggestion
- `static`: build HDF5 C library from source (see below for a list of options)
```

hdf5-sys/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+The following features affect the HDF5 build options when compiling it from source:
+
+- `mpio`: enable MPI support
```
aldanor, line 26: @magnusuMET Should this be removed as we no longer support mpio/static together?

hdf5-sys/README.md
```diff
@@ -0,0 +1,32 @@
...
+As of the time of writing, the version of the HDF5 library that is built is 1.10.5,
+but it may be incremented later.
+
+## Crate features
+
+Features propagated upwards will be detected based on features available for the chosen library. The library selected can always be overridden by specifying the environment variable `HDF5_DIR`.
+
+General features:
+
+- `mpio`: enable MPI support (in static mode, also enable it in the HDF5 C library)
```
aldanor, line 21:
```suggestion
- `mpio`: enable MPI support (not supported in static mode)
```

aldanor:
@magnusuMET I just got to re-reading this PR again, I think it's 99% done, just a few docs tweaks – because you removed mpio + static support it makes sense to fix the docs as well.

GH Actions look nice - I'll try to completely remove Travis/AppVeyor after this is merged so as GHA is used for all testing.

Profile-specific debug/release mode looks nice. I can't test it right now but I hope that the naming hacks actually work 👍 (should we maybe test it on GHA in both debug/release modes? idk)

Finally, what about hdf5-derive, it can't be tested on GHA with the static mode I guess?

aldanor:
There's also a weird macos failure on Travis, hm... (but seems unrelated at first glance)

aldanor:
> There's also a weird macos failure on Travis, hm... (but seems unrelated at first glance)

Yea. The minimum available version on conda is now 1.8.18 and not 1.8.9 - I'll fix it after it's merged, as part of porting to GHA.

hdf5-sys/README.md
magnusuMET: I added a note regarding mutual exclusivity in `hdf5-sys/README.md`. We could add a note there that this restriction could be lifted.

magnusuMET:
@aldanor Thanks again for taking interest in this PR!

I've issued a build on my GA for testing the `--release` on all platforms (https://github.com/magnusuMET/hdf5-rust/runs/910222245?check_suite_focus=true). Seems a bit superfluous to test this for every build, but it is trivial to add. This would add maybe 10 minutes of build time to the CI (windows CI is slow).

`hdf5-derive` can be tested if we could remove `hdf5` from dev-dependencies. The `hdf5` that the tests are running against is not the same as what we are trying to build.

aldanor:
> hdf5-derive can be tested if we could remove hdf5 from dev-dependencies. The hdf5 that the tests are running against is not the same as what we are trying to build.

@magnusuMET In reality, there's no reason whatsoever that hdf5-derive won't work with static build, after all it's just compile-fail tests that involve hdf5 itself...

As for release/debug builds, I'll take a look at that once again I port everything to GHA and fix macos build errors. Just to make sure we have good coverage of hdf5 installation options (system installs, homebrew, apt, conda, static) and versions (various 1.8 and 1.10, and soon 1.12 as well), plus release/debug.

This being said, I think it looks LGTM. After you confirm everything's done I'll merge it and will try to release a version asap.

hdf5-sys/README.md
aldanor: @magnusuMET Yea, but "mpio" version is no longer supported here technically, no (as of this moment)?

Could add a separate note below that we don't support mpio for static build but it could be lifted/implemented if really needed.

magnusuMET: `mpio` is supported for non-static builds, this feature was there before this PR. I don't think we have CI for this.

magnusuMET:
@aldanor There is a cyclical dependency between `hdf5` and `hdf5-derive`. The tests could be moved to `hdf5` which resolves this.

magnusuMET:
@aldanor I believe we are ready to merge!

hdf5-sys/README.md
aldanor: @magnusuMET That's not what I mean :) Currently we have:

```markdown
The following features affect the HDF5 build options when compiling it from source:

- `mpio`: enable MPI support
```

Which is not true because we don't support mpio when compiling it from source? (the `mpio` notice above as a "general feature" makes sense of course, we don't test it on CI but it works for normal dynamic builds)

magnusuMET: Now I see what you mean! Thanks for the patience :+1:

magnusuMET: @aldanor Forgot to mention you..

aldanor:
Merged, thanks @magnusuMET for good work :) I'll look into migrating CI to GHA later today (also docs pushing and all that), hoping to release a next version with all this soon.

gauteh:
Thanks everyone! This makes my life significantly easier!!

PR #75, opened 2020-02-24
hdf5-sys: Add HDF5 Optimized Functions

kalcutter:


aldanor:
Thanks for the PR - interesting, how did I miss those. Are these part of the normal libhdf5 (and not libhdf5_hl), correct?

aldanor:
https://support.hdfgroup.org/HDF5/doc/HL/RM_HDF5Optimized.html

Indeed, it seems like it's part of `libhdf5_hl`...

(Which means: if we do it, it should be done properly, i.e. either pull it all out into a separate crate or, at the very least, provide proper linking to `libhdf5_hl.so` on all platforms?)

kalcutter:
Ahh. Sorry I didn't realize that these were in a separate library. Do you think its better to make another crate for the HL functions? It would be easier to put them in this crate (maybe behind a cfg gate)? Putting them in a separate crate would be more modular I suppose.

kalcutter:
Thinking about this more... A new crate probably makes the most sense (hdf5-hl-sys?).

magnusuMET:
In the static build (#67) I've combined the two libraries, with a feature flag to select "hl"

kalcutter:
Something similar here does seem like the most pragmatic solution. Then we could directly reuse the code in build.rs (and other places). It would be less work/duplication.

magnusuMET:
Would certainly be less problematic than trying to find compatible libraries across different versions of `hdf5-sys` and a `hdf5-hl-sys`

kalcutter:
Closed in favor of #77.

PR #78, opened 2020-03-02
hdf5-sys: Fix wrong H5Z_filter_t type

kalcutter:
H5Z_filter_t has type int not hid_t.

### DIFF
--- a/hdf5-sys/src/h5z.rs
+++ b/hdf5-sys/src/h5z.rs
@@ -6,19 +6,21 @@ pub use self::H5Z_cb_return_t::*;
 
 use crate::internal_prelude::*;
 
-pub const H5Z_FILTER_ERROR: hid_t = -1;
-pub const H5Z_FILTER_NONE: hid_t = 0;
-pub const H5Z_FILTER_DEFLATE: hid_t = 1;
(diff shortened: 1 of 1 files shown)


kalcutter:
This fixes `H5Zregister` and friends when building against hdf5 1.10.x which was broken because of this type mismatch.

aldanor:
Well... https://github.com/aldanor/hdf5-rust/commit/11990985bd9dd17b6cb63a409c7d03537fadd8b5

Maybe just cherry-pick that one?

I really need to finish merging that mega-branch :(

kalcutter:
Can you make a new bugfix release without the rest of the mega-branch? It would be helpful for the people using an official crates.io version.

aldanor:
Good point, I can probably pull the first 30-ish commits or so from there - most of them being dependency version bumps, compatibility fixes, bugfixes like this one etc.

aldanor:
@kalcutter I've pushed #79 with all changes unrelated to the mega-branch itself (including this fix). Something's failing there (because there's tons of stuff) but I don't have time to look at it in detail right now, looks like some obscure ndarray failures on appveyor :/

PR #89, opened 2020-04-04
add resizable_idx to DatasetBuilder

### LINKED ISSUE #88: Can't create resizable datasets for only some indices
I am trying to create a dataset where only one of the indices are unlimited. The methods exposed on `DatasetBuilder` only allows all or none of the indices to be resizable.

mulimoen:
An idea for how to set certain indices to be resizable

Fixes #88

### DIFF
--- a/src/hl/container.rs
+++ b/src/hl/container.rs
@@ -127,7 +127,7 @@ impl<'a> Reader<'a> {
                 );
             }
 
-            let mspace = Dataspace::try_new(&out_shape, false)?;
+            let mspace = Dataspace::try_new(&out_shape, None)?;
             let size = out_shape.iter().product();
             let mut vec = Vec::with_capacity(size);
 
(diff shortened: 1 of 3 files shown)


aldanor:
@mulimoen Ok, this one will have to wait until the dcpl / dataset rework I think, because the entire dataset builder got thrown away and replaced.

Also, I implemented full extents api there, which I think should cover this case as well? See the tests: https://github.com/aldanor/hdf5-rust/blob/feature/dcpl/src/hl/extents.rs#L488

mulimoen:
@aldanor No problem with closing this, it was currently just a problem for now that extents were either all or nothing. The new extents API would be perfect for my use case (leaving last dimension resizeable).

aldanor:
Ok, great then! I'll close this for now to reduce the number of open prs. I wanted to release a new (non-breaking-change) version this weekend if possible with static builds / gha / 1.12 / alloc fixes etc, then I'll jump straight to finishing feature/dcpl with the whole new dataset api and all that (breaking change).

PR #92, opened 2020-04-23
Implement Dimension for slices

phil-opp:
Add a `?Sized` bound to the `Dimension` implementation for `&T` to ensure that unsized slices can be used as shape.

I also added a small `slice_as_shape` test function that only compiles with the `?Sized` bound.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Added
+- Slices can now be used used where trait item `Dimension` is required.
+
 ### Changed
 - Dependencies are bumped to newest major versions; `ndarray` users may now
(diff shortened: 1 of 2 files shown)


phil-opp:
Rebased to run PR on new Github Actions based CI.

mulimoen:
This makes sense to include in our next release. Rebased to check CI

aldanor:
This wasn't merged in because I think it will be thrown out once we merge dcpl branch (completely superceded by Extents and all that).

mulimoen:
I think a `0.7.1` change with these minor conveniences makes sense. The `dcpl` change is rather major and will require a major bump.

aldanor:
If this is treated as a (perhaps) temporary convenience change then it's probably ok, yea.

PR #95, opened 2020-05-03
Add 1.12.0 support

### LINKED ISSUE #94: support hdf5 v1.12
Building against hdf5 1.12 fails because Version::parse regards `H5_VERSION 1.12.0` as invalid. Are there any breaking changes that prevent this crate from being used with 1.12?

mulimoen:
Add the newest changes according to https://portal.hdfgroup.org/display/HDF5/Software+Changes+from+Release+to+Release+for+HDF5-1.12. Currently compiles and tests complete sucessfully on all platforms but new functions are not tested nor used anywhere.

Fixes #94

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,31 @@ jobs:
           command: ${{matrix.command}}
           args: "${{matrix.command == 'fmt' && '--all -- --check' || '-- -D warnings'}}"
 
+  brew:
+    name: brew
+    runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - {version: hdf5@1.8}
+          - {version: hdf5@1.10}
+          - {version: hdf5@1.12}
+          - {version: hdf5-mpi, mpi: true}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with: {submodules: true}
+      - name: Install Rust (${{matrix.rust}})
+        uses: actions-rs/toolchain@v1
+        with: {toolchain: stable, profile: minimal, override: true}
+      - name: Install HDF5 (${{matrix.version}})
+        run: brew install ${{matrix.version}}
+      - name: Build and test all crates
+        run: |
+          [ "${{matrix.mpi}}" != "" ] && FEATURES=mpio
+          cargo test -vv --features="$FEATURES"
+
   conda:
     name: conda
     runs-on: ${{matrix.os}}-latest
@@ -55,6 +80,9 @@ jobs:
           - {os: macos, version: 1.10.5, mpi: openmpi, channel: conda-forge, rust: beta}
           - {os: ubuntu, version: 1.10.6, channel: anaconda, rust: stable}
           - {os: ubuntu, version: 1.10.6, mpi: mpich, channel: conda-forge, rust: nightly}
+          - {os: ubuntu, version: 1.12.0, mpi: openmpi, channel: conda-forge, rust: stable}
+          - {os: macos, version: 1.12.0, channel: conda-forge, rust: stable}
+          - {os: windows, version: 1.12.0, channel: conda-forge, rust: stable}
     defaults:
       run:
         shell: bash -l {0}
@@ -101,26 +129,6 @@ jobs:
         with: {toolchain: '${{matrix.rust}}', profile: minimal, override: true}
       - name: Build and test all crates
         run: cargo test --workspace -v --features hdf5-sys/static,hdf5-sys/zlib --exclude hdf5-derive
-
-  brew:
-    name: brew
-    runs-on: macos-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        version: [1.8, '1.10']
-        rust: [stable]
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v2
-        with: {submodules: true}
-      - name: Install Rust (${{matrix.rust}})
-        uses: actions-rs/toolchain@v1
-        with: {toolchain: '${{matrix.rust}}', profile: minimal, override: true}
-      - name: Install HDF5 (${{matrix.version}})
-        run: brew install hdf5@${{matrix.version}}
-      - name: Build and test all crates
-        run: cargo test -v
 
   apt:
     name: apt
--- a/hdf5-sys/build.rs
+++ b/hdf5-sys/build.rs
@@ -300,10 +300,14 @@ mod macos {
         }
         // We have to explicitly support homebrew since the HDF5 bottle isn't
         // packaged with pkg-config metadata.
-        let (v18, v110) = if let Some(version) = config.version {
-            (version.major == 1 && version.minor == 8, version.major == 1 && version.minor == 10)
+        let (v18, v110, v112) = if let Some(version) = config.version {
+            (
+                version.major == 1 && version.minor == 8,
+                version.major == 1 && version.minor == 10,
+                version.major == 1 && version.minor == 12,
+            )
         } else {
-            (false, false)
+            (false, false, false)
         };
         println!(
             "Attempting to find HDF5 via Homebrew ({})...",
@@ -311,19 +315,35 @@ mod macos {
                 "1.8.*"
             } else if v110 {
                 "1.10.*"
+            } else if v112 {
+                "1.12.*"
             } else {
                 "any version"
             }
         );
-        if !v18 {
+        if !(v18 || v110) {
+            if let Some(out) = run_command("brew", &["--prefix", "hdf5@1.12"]) {
+                if is_root_dir(&out) {
+                    config.inc_dir = Some(PathBuf::from(out).join("include"));
+                }
+            }
+        }
+        if config.inc_dir.is_none() && !v18 {
             if let Some(out) = run_command("brew", &["--prefix", "hdf5@1.10"]) {
                 if is_root_dir(&out) {
                     config.inc_dir = Some(PathBuf::from(out).join("include"));
                 }
             }
         }
-        if config.inc_dir.is_none() && !v110 {
+        if config.inc_dir.is_none() {
             if let Some(out) = run_command("brew", &["--prefix", "hdf5@1.8"]) {
+                if is_root_dir(&out) {
+                    config.inc_dir = Some(PathBuf::from(out).join("include"));
+                }
+            }
+        }
+        if config.inc_dir.is_none() {
+            if let Some(out) = run_command("brew", &["--prefix", "hdf5-mpi"]) {
                 if is_root_dir(&out) {
                     config.inc_dir = Some(PathBuf::from(out).join("include"));
                 }
--- a/hdf5-sys/build.rs
+++ b/hdf5-sys/build.rs
@@ -25,7 +25,7 @@ impl Version {
     }
 
     pub fn parse(s: &str) -> Option<Self> {
-        let re = Regex::new(r"^(1)\.(8|10)\.(\d\d?)(_\d+)?(-patch\d+)?$").ok()?;
+        let re = Regex::new(r"^(1)\.(8|10|12)\.(\d\d?)(_\d+)?(-patch\d+)?$").ok()?;
         let captures = re.captures(s)?;
         Some(Self {
             major: captures.get(1).and_then(|c| c.as_str().parse::<u8>().ok())?,
@@ -35,7 +35,10 @@ impl Version {
     }
 
     pub fn is_valid(self) -> bool {
-        self.major == 1 && ((self.minor == 8 && self.micro >= 4) || (self.minor == 10))
+        self.major == 1
+            && ((self.minor == 8 && self.micro >= 4)
+                || (self.minor == 10)
+                || (self.minor == 12 && self.micro == 0))
     }
 }
 
@@ -592,6 +595,7 @@ impl Config {
         assert!(version >= Version::new(1, 8, 4), "required HDF5 version: >=1.8.4");
         let mut vs: Vec<_> = (5..=21).map(|v| Version::new(1, 8, v)).collect(); // 1.8.[5-21]
         vs.extend((0..=5).map(|v| Version::new(1, 10, v))); // 1.10.[0-5]
+        vs.push(Version::new(1, 12, 0)); // 1.12.0
         for v in vs.into_iter().filter(|&v| version >= v) {
             println!("cargo:rustc-cfg=hdf5_{}_{}_{}", v.major, v.minor, v.micro);
             println!("cargo:version_{}_{}_{}=1", v.major, v.minor, v.micro);
--- a/hdf5-sys/src/h5i.rs
+++ b/hdf5-sys/src/h5i.rs
@@ -8,20 +8,27 @@ pub enum H5I_type_t {
     H5I_UNINIT = -2,
     H5I_BADID = -1,
     H5I_FILE = 1,
-    H5I_GROUP = 2,
-    H5I_DATATYPE = 3,
-    H5I_DATASPACE = 4,
-    H5I_DATASET = 5,
-    H5I_ATTR = 6,
+    H5I_GROUP,
+    H5I_DATATYPE,
+    H5I_DATASPACE,
+    H5I_DATASET,
+    #[cfg(hdf5_1_12_0)]
+    H5I_MAP,
+    H5I_ATTR,
+    #[cfg(not(hdf5_1_12_0))]
     #[cfg_attr(hdf5_1_10_2, deprecated(note = "deprecated in HDF5 1.10.2"))]
-    H5I_REFERENCE = 7,
-    H5I_VFL = 8,
-    H5I_GENPROP_CLS = 9,
-    H5I_GENPROP_LST = 10,
-    H5I_ERROR_CLASS = 11,
-    H5I_ERROR_MSG = 12,
-    H5I_ERROR_STACK = 13,
-    H5I_NTYPES = 14,
+    H5I_REFERENCE,
+    H5I_VFL,
+    #[cfg(hdf5_1_12_0)]
+    H5I_VOL,
+    H5I_GENPROP_CLS,
+    H5I_GENPROP_LST,
+    H5I_ERROR_CLASS,
+    H5I_ERROR_MSG,
+    H5I_ERROR_STACK,
+    #[cfg(hdf5_1_12_0)]
+    H5I_SPACE_SEL_ITER,
+    H5I_NTYPES,
 }
 
 #[cfg(hdf5_1_10_0)]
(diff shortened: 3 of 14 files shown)


hdf5-sys/src/h5i.rs
```diff
@@ -8,20 +8,27 @@ pub enum H5I_type_t {
...
-    H5I_GROUP = 2,
-    H5I_DATATYPE = 3,
-    H5I_DATASPACE = 4,
-    H5I_DATASET = 5,
-    H5I_ATTR = 6,
+    H5I_GROUP,
+    H5I_DATATYPE,
+    H5I_DATASPACE,
+    H5I_DATASET,
+    #[cfg(hdf5_1_12_0)]
```
aldanor, line 15: This all starts to look fairly fragile, did they renumber existing enums? 🤦

mulimoen: They added and removed members here. This caused some very odd errors that was hard to find...

aldanor:
Ok, I'll look at this now, will get back with some comments.

Given the sizeable change list, I'll probably have to diff the C headers myself, just to give it a second look, manually.

The dodgy parts are the whole func1/func2 (which are macroed in C code) as they can cause linkage errors if not done carefully, and the whole enum renumbering business (does it mean it's a breaking change now or not?).

Also, regarding the CI:
- we also want to build it on macos (simply `@1.12` for brew builds)
- for homebrew, there's now also `hdf5-mpi`, only available as 1.12 (`hdf5-mpi@1.12`). It uses `openmpi` as the MPI backend
- would be nice to stratify the builds between different rust toolchains and different conda build types (openmpi/mpich/nompi) as you can see I've tried to do to get max coverage
- Note that while I have 40 concurrent jobs limit on my account, the number of concurrent macOS jobs is = 5, and we will end up having something like 7. It may make sense then to schedule brew jobs at the very start as they seem to be the fastest to finish anyway.

mulimoen:
I'll add some more brew jobs! It seems the HDF group restricted binary releases for `HDF5` so we need to log in the get these it seems.

The func1/func2 is a bit annoying. The original names will be shadowed based on API compatability version selected during build time. We can't just point `func` to point to either `func1` or `func2` based on the version number. We can point `func1` and `func` to the same function for versions before the introduction of `func2` to avoid having a breaking change.

aldanor:
> It seems the HDF group restricted binary releases for `HDF5` so we need to log in the get these it seems.

I've raised this on their forum, please feel free to chime in :) https://forum.hdfgroup.org/t/please-provide-stable-urls-for-official-binaries-without-auth/7378

mulimoen:
I went through and manually diffed the HPXpublic.h headers, which resulted in some additional changes to some additional enums.

hdf5-sys/src/h5r.rs
```diff
@@ -46,3 +60,49 @@ extern "C" {
...
 pub use self::H5Rdereference1 as H5Rdereference;
+
+#[cfg(hdf5_1_12_0)]
+pub const H5R_REF_BUF_SIZE: usize = 64;
+
+#[cfg(hdf5_1_12_0)]
+#[repr(C)]
+pub union H5R_ref_t_u {
+    __data: [u8; H5R_REF_BUF_SIZE],
+    align: i64,
```
aldanor, line 71: `Default` impl / zeroed?

hdf5-sys/src/h5r.rs
```diff
@@ -46,3 +60,49 @@ extern "C" {
...
+}
+
+#[cfg(hdf5_1_12_0)]
+#[repr(C)]
+pub struct H5R_ref_t {
+    u: H5R_ref_t_u,
+}
```
aldanor, line 78: `Default` impl / zeroed?

hdf5-sys/build.rs
```diff
@@ -297,35 +300,55 @@ mod macos {
...
             } else if v110 {
                 "1.10.*"
+            } else if v112 {
+                "1.12.*"
             } else {
                 "any version"
             }
         );
-        if !v18 {
+        if !(v18 || v110) {
```
aldanor, line 324: There's tons of copy/paste now and weird logic in the whole homebrew section of this, should probably be rewritten so it's cleaner... (also `hdf5-mpi` is only available for 1.12 and later).

(could be done separately outside of this PR, just noting)

.github/workflows/ci.yml
```diff
@@ -55,6 +75,9 @@ jobs:
           - {os: macos, version: 1.10.5, mpi: openmpi, channel: conda-forge, rust: beta}
           - {os: ubuntu, version: 1.10.6, channel: anaconda, rust: stable}
           - {os: ubuntu, version: 1.10.6, mpi: mpich, channel: conda-forge, rust: nightly}
+          - {os: ubuntu, version: 1.12.0, channel: conda-forge, rust: stable}
```
aldanor, line 78: Maybe use an mpi version here on ubuntu?

.github/workflows/ci.yml
```diff
@@ -35,6 +35,26 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with: {submodules: true}
+      - name: Install Rust (${{matrix.rust}})
+        uses: actions-rs/toolchain@v1
+        with: {toolchain: '${{matrix.rust}}', profile: minimal, override: true}
+      - name: Install HDF5 (${{matrix.hdf5}})
+        run: brew install ${{matrix.hdf5}}
+      - name: Build and test all crates
+        run: cargo test -vv
```
aldanor, line 56: This is not actually enabling an MPI build as you have to pass in `--features` (see like how it's done below)

aldanor:
@mulimoen I've left a few comments, most of them cosmetic/minor.

One concern here is this: we will now import things like `H5Literate1` in `hdf5::hl::group` which, I think, will cause `deprecated` warnings for downstream builds on 1.12 (correct me if I'm wrong or missing something).

src/hl/group.rs
```diff
@@ -6,8 +6,8 @@ use hdf5_sys::{
     h5d::H5Dopen2,
     h5g::{H5G_info_t, H5Gcreate2, H5Gget_info, H5Gopen2},
     h5l::{
-        H5L_info_t, H5L_iterate_t, H5Lcreate_hard, H5Lcreate_soft, H5Ldelete, H5Lexists,
-        H5Literate, H5Lmove, H5L_SAME_LOC,
+        H5L_info1_t as H5L_info_t, H5L_iterate1_t as H5L_iterate_t, H5Lcreate_hard, H5Lcreate_soft,
```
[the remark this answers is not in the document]
mulimoen: Maybe we should use the v2 variants when these are available?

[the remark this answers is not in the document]
aldanor: I think we have to, but then you'll end up with lots of `cfg_attr` boilerplate in hdf5 crate code. I was even thinking of some helper macros here or aliases, will it help? Hm.

hdf5-sys/src/h5o.rs
```diff
@@ -318,3 +340,97 @@ extern "C" {
         op: H5O_iterate_t, op_data: *mut c_void, lapl_id: hid_t,
     ) -> herr_t;
 }
+
+pub const H5O_MAX_TOKEN_SIZE: usize = 16;
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct H5O_token_t {
+    __data: [u8; H5O_MAX_TOKEN_SIZE],
```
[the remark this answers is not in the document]
mulimoen: I'm going to use H5O_TOKEN_UNDEF here

hdf5-sys/src/h5o.rs
```diff
@@ -318,3 +340,97 @@ extern "C" {
...
+        op: H5O_iterate2_t, op_data: *mut c_void, fields: c_uint, lapl_id: hid_t,
+    ) -> herr_t;
+}
+
+#[cfg(hdf5_1_12_0)]
+pub use self::globals::*;
+
+#[cfg(all(not(all(target_env = "msvc", not(feature = "static"))), hdf5_1_12_0))]
+mod globals {
+    pub use crate::h5i::hid_t as id_t;
```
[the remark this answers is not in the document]
mulimoen: the `id_t` is part of the extern static macro

hdf5-sys/build.rs
mulimoen: #112

src/hl/group.rs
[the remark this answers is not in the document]
mulimoen: We will have to do this on a function-to-function basis I think. Since the topcrate is higher level than `hdf5-sys`, we can probably hide the different versions inside each function.

.github/workflows/ci.yml
```diff
@@ -35,6 +35,28 @@ jobs:
           command: ${{matrix.command}}
           args: "${{matrix.command == 'fmt' && '--all -- --check' || '-- -D warnings'}}"
 
+  brew:
+    name: brew
+    runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        hdf5: [hdf5@1.8, hdf5@1.10, hdf5@1.12, hdf5-mpi]
```
aldanor, line 44: Maybe like this?

```yaml
rust: [stable]
include:
  - {version: 1.8}
  - {version: '1.10'}
  - {version: 1.12}
  - {version: 1.12, mpi: true}
```

And then like

```sh
[ "${{matrix.mpi}}" == "" ] && PKG=hdf || PKG=hdf5-mpi
brew install $PKG@${{matrix.version}}
```

hdf5-sys/src/h5l.rs
```diff
@@ -182,32 +226,79 @@ extern "C" {
...
+    #[cfg_attr(not(hdf5_1_12_0), link_name = "H5Lget_info")]
+    pub fn H5Lget_info1(
+        loc_id: hid_t, name: *const c_char, linfo: *mut H5L_info1_t, lapl_id: hid_t,
     ) -> herr_t;
+    #[cfg(hdf5_1_12_0)]
+    pub fn H5Lget_info2(
+        loc_id: hid_t, name: *const c_char, linfo: *mut H5L_info2_t, lapl_id: hid_t,
+    ) -> herr_t;
+    #[cfg_attr(
+        hdf5_1_12,
```
aldanor, line 239: @mulimoen Should this be `hdf5_1_12_0`? (here and in tons of other places)

.github/workflows/ci.yml
```diff
@@ -35,6 +35,34 @@ jobs:
...
+          - {version: 1.12}
+          - {version: 1.12, mpi: true}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with: {submodules: true}
+      - name: Install Rust (${{matrix.rust}})
+        uses: actions-rs/toolchain@v1
+        with: {toolchain: '${{matrix.rust}}', profile: minimal, override: true}
+      - name: Install HDF5 (${{matrix.hdf5}})
```
aldanor, line 57: `matrix.version`?

mulimoen: Will fix. Feel free to cancel the actions

hdf5-sys/src/h5l.rs
mulimoen: Well spotted!

hdf5-sys/src/h5r.rs
```diff
@@ -46,3 +60,49 @@ extern "C" {
...
+#[cfg(hdf5_1_12_0)]
+#[repr(C)]
+pub union H5R_ref_t_u {
+    __data: [u8; H5R_REF_BUF_SIZE],
+    align: i64,
+}
+
+#[cfg(hdf5_1_12_0)]
+#[repr(C)]
+pub struct H5R_ref_t {
```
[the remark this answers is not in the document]
mulimoen: How should we format the union in the debug implementation?

mulimoen:
> import things like `H5Literate1` in `hdf5::hl::group` which, I think, will cause `deprecated` warnings for downstream builds on 1.12

Unfortunately this breaks compile time tests, so I've set these to be ignored. I'll open an issue for dealing with deprecated items.

mulimoen:
It was trivial to use v2 of the API...

src/hl/group.rs
```diff
@@ -1,17 +1,24 @@
...
-    h5l::{
-        H5L_info1_t as H5L_info_t, H5L_iterate1_t as H5L_iterate_t, H5Lcreate_hard, H5Lcreate_soft,
-        H5Ldelete, H5Lexists, H5Literate1 as H5Literate, H5Lmove, H5L_SAME_LOC,
-    },
+    h5l::{H5Lcreate_hard, H5Lcreate_soft, H5Ldelete, H5Lexists, H5Lmove, H5L_SAME_LOC},
     h5p::{H5Pcreate, H5Pset_create_intermediate_group},
 };
 
+#[cfg(not(hdf5_1_12_0))]
+use hdf5_sys::h5l::{
```
aldanor, line 14: Hmm, so that's what I was mentioning earlier, maybe we're doing it all backwards? 

```rust
    #[cfg_attr(hdf5_1_12_0, deprecated(note = "deprecated in HDF5 1.12.0, use H5Literate2()"))]
    #[cfg_attr(not(hdf5_1_12_0), link_name = "H5Literate")]
    pub fn H5Literate1(
        grp_id: hid_t, idx_type: H5_index_t, order: H5_iter_order_t, idx: *mut hsize_t,
        op: H5L_iterate1_t, op_data: *mut c_void,
    ) -> herr_t;
    #[cfg(hdf5_1_12_0)]
    pub fn H5Literate2(
        grp_id: hid_t, idx_type: H5_index_t, order: H5_iter_order_t, idx: *mut hsize_t,
        op: H5L_iterate2_t, op_data: *mut c_void,
    ) -> herr_t;

    // Mimic HDF5 C header macros?
    #[cfg(not(hdf5_1_12_0))]
    pub use self::H5Literate1 as H5Literate;
    #[cfg(hdf5_1_12_0)]
    pub use self::H5Literate2 as H5Literate;
```

Then in `hdf5::hl::group` you just `use hdf5_sys::h5l::H5Literate`?

mulimoen: The only confusing bit would be that C users might expect a different version, as the macro conditionally uses a particular version, which might not be the latest one

aldanor: Hm?

So, pre-1.12:

- If you use `H5Literate()`, it would resolve to the same in Rust/C
- We also have `H5Literate1` which resolves to the `H5Literate()`
- `H5Literate2()` doesn't exist in either

Post 1.12:

- `H5Literate2()` resolves to the same thing in Rust/C
- `H5Literate1()` resolves to the same thing in Rust/C
- We also have `H5Literate` which resolves to `H5Literate2()`

mulimoen: A crate could directly use `H5Literate1` and remain forwards compatible, although with deprecation warnings. It would probably be a good idea to skim through the crate to make sure we have such aliasing for all structs and functions. This can be left for a future PR

mulimoen: The `H5version.h` crate is responsible for selecting the API version of each function based on compile time flags, but we bypass this entirely and always use the latest API.

aldanor: Yea, right. I don't think there's an easy way out and in fact what we're doing is probably a clearer way to go as long as we do it with all 1/2/3 versioned structs and functions, tag them all with deprecation warnings and provide aliases to select one automatically like the one you added. As you've said, probably a separate PR to revise all existing versioned bindings, mind opening a ticket?

mulimoen: #114

aldanor:
Skimming through changes again, I *think* it looks good to go, we'll fix up the opened tickets later. Any thoughts?

Would probably make sense to release a version too to include all the recent changes.

hdf5-sys/src/h5r.rs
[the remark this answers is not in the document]
aldanor: You can't derive Debug for unions and since they're untagged it's typically ok to leave them without Debug I think.

mulimoen:
@aldanor I rebased and fixed two renamings, so we are compatible on minor semver.

It would be great to have a new version cut, both to test `docs.rs`, and to make it easier to build `hdf5` on all platforms!

hdf5-sys/src/h5p.rs
```diff
@@ -656,7 +688,13 @@ extern "C" {
     pub fn H5Pset_virtual_view(plist_id: hid_t, view: H5D_vds_view_t) -> herr_t;
     pub fn H5Pget_chunk_opts(plist_id: hid_t, opts: *mut c_uint) -> herr_t;
     pub fn H5Pset_chunk_opts(plist_id: hid_t, opts: c_uint) -> herr_t;
-    pub fn H5Pencode(plist_id: hid_t, buf: *mut c_void, nalloc: *mut size_t) -> herr_t;
+    #[cfg_attr(hdf5_1_12_0, deprecated(note = "deprecated in HDF5 1.12.0, use H5Pencode2"))]
```
aldanor, line 691:
```suggestion
    #[cfg_attr(hdf5_1_12_0, deprecated(note = "deprecated in HDF5 1.12.0, use H5Pencode2()"))]
```

hdf5-sys/src/h5s.rs
```diff
@@ -53,7 +56,13 @@ extern "C" {
     ) -> herr_t;
     pub fn H5Scopy(space_id: hid_t) -> hid_t;
     pub fn H5Sclose(space_id: hid_t) -> herr_t;
-    pub fn H5Sencode(obj_id: hid_t, buf: *mut c_void, nalloc: *mut size_t) -> herr_t;
+    #[cfg_attr(hdf5_1_12_0, deprecated(note = "deprecated in HDF5 1.12.0, use H5Sencode2"))]
```
aldanor, line 59:
```suggestion
    #[cfg_attr(hdf5_1_12_0, deprecated(note = "deprecated in HDF5 1.12.0, use H5Sencode2()"))]
```

aldanor:
Cool! Just two very minor deprecated message fixes (for consistency throughout), and I'll merge this in then.

I'll get to pushing a release then - I was thinking 0.7.0, just because of the amount of stuff, changes to the build system and a new major HDF5 version.

mulimoen:
And fixed!

gauteh:
It seems hdf5-src is still tagged to 1.10, any reason for that?

aldanor:
I really think we should support multiple major versions for static builds

gauteh:
Yes, that would be great. Seems to be significant breakage between hdf5
versions. So would be great to pin to an older version of hdf-src.

fre. 28. aug. 2020, 20:12 skrev Ivan Smirnov <notifications@github.com>:

aldanor:
Problem is, if we ship 1.8/1.10/1.12/1.13, we'd have to include source code for all those HDF5 versions, thus quadrupling the size which is already substantial.

Idk, maybe we can include zipped source in the crate and unpack it at build time?

mulimoen:
@gauteh hdf5-src is tagged to 1.10 as that should be well supported by 'netCDF', which is why this was implemented in the first place

mulimoen:
@aldanor we could have a base version, and a zipped diff for every other version released. This should not be too massive

aldanor:
@mulimoen Source can be compressed 5-6x, so I think even if we provide 4 packed versions it will be smaller than the current crate size

gauteh:
Ivan Smirnov writes on August 28, 2020 20:39:
> @mulimoen Source can be compressed 5-6x, so I think even if we provide 4 packed versions it will be smaller than the current crate size

Is this for crates.io? Or because you need more submodules?

aldanor:
For crates.io; and the easiest way would probably be to add more submodules, like `ext/1.8`, `ext/1.10`, `ext/1.12`, etc...

PR #103, opened 2020-07-28
Configure the docs.rs build

magnusuMET:
By setting metadata in `Cargo.toml` (https://docs.rs/about#metadata), where we build hdf5 statically, we get documentation for "free" on `docs.rs`. We no longer have to build and deploy the documentation ourselfs.

@aldanor This would be a great PR to merge before the next release!


aldanor:
Haha, I have exactly the same chance in my current gha branch, I guess I can just rebase it then :)

// Mind replacing those double line breaks with single ones?

magnusuMET:
@aldanor Done!

Issue #104, opened 2020-07-30
Reproducible test_dataset crash on Windows / 1.10.2 from anaconda


aldanor:
https://github.com/aldanor/hdf5-rust/runs/924910831

Anyone who has a windows box is welcome to check this out ^

Seems to results in `STATUS_HEAP_CORRUPTION` in `test_dataset`.

aldanor:
@magnusuMET ^ btw: this will be a blocker for migrating everything fully to GHA and releasing a new version, I think.

There's another reproducible macOS test failure on 1.8.18 with homebrew build, but that one I'll take on myself since I have a mac box at home.

magnusuMET:
I can reproduce on a virtual machine running on linux. The error occurs in 
`test_read_write_varlen_struct` -> `test_read` -> `assert_eq` (line 113)

The error is probably due to some undefined behaviour. Moving the `?` operator to
```rust
let out_vec = ds.read_raw::<T>()?;
```
makes the assertation pass, but it again fails on the next assertation.

aldanor:
Hm, interesting, needs further investigation.

What we need to figure out, I guess - is it our fault or a known HDF5 bug? If it's the first one causing some UB (although it looks like some very reproducible UB), we have to fix it; if it's a known bug - just skip this test, or maybe try to work around it.

aldanor:
I was thinking also to schedule a weekly run of ~200-250 jobs, running a full test matrix over all known HDF5 versions from all sources on all platforms and all Rust versions - this may help uncover more crash bugs in the future.

(Will probably do it later after this crashbug is investigated and GHA PR is merged in)

magnusuMET:
@aldanor I have a question regarding memory managment of the string types. When reading raw values from a dataset, who has the responsibility for calling `free` on the allocated strings? If it is us, which I assume since the variable strings have a `Drop` implementation, are we sure we are using the same allocator to free as was used to allocate the strings in `hdf5-c`? `STATUS_HEAP_CORRUPTION` looks like an allocator error

magnusuMET:
See https://portal.hdfgroup.org/display/HDF5/H5_FREE_MEMORY,

> It is especially important to use this function to free memory allocated by the library on Windows. The C standard library is implemented in dynamic link libraries (DLLs) known as the C run-time (CRT). Each version of Visual Studio comes with two CRT DLLs (debug and release) and allocating and freeing across DLL boundaries can cause resource leaks and subtle bugs due to heap corruption.
[...]

PR #105, opened 2020-08-02
Switch to GitHub Actions, add MPI tests

aldanor:
- Switch all CI testing entirely to GitHub Actions
- Expand testing matrix, now 24 various builds are running (under 10m runtime total)
- Run scheduled daily tests to see if anything breaks under nightly
- Remove Travis / AppVeyor configs
- Update README badges and some outdated info, add link to docs.rs (next release if it works out)
- Actually test `--features=mpio`, apt and conda-forge, OpenMPI and MPICH (that took a while to figure out, particularly with conda one)
- Fix a bunch of clippy errors; cargo clippy and cargo fmt also run on CI

This is currently blocked by #104 (reproducible crash on Windows) – if anyone with a windows box wants to help out, yw. If it turns out to be a known bug or something, we can mention it in readme and exclude it from CI.


magnusuMET:
It would be preferable to have the `lints` job on top of the list of jobs to run. It might also be a good idea to run `cargo doc` to ensure documentation is kept up to date.

magnusuMET:
We could split the crater runs from the regular CI by creating a separate `.yml` that only runs as a `cron` job.

aldanor:
> We could split the crater runs from the regular CI by creating a separate `.yml` that only runs as a `cron` job.

Yes, that was my plan. A bit unfortunate that a lot will have to be copied/pasted, of course.

aldanor:
@magnusuMET Moved lints up, there's not much to check now really, but I have big docs work/rework in mind later this year (especially if we manage to build docs.rs docs), will revise it then.

This whole endeavour seems to have been well worth it since it allowed us to discover and fix #104 :)

PR #107, opened 2020-08-04
Use hdf5 allocator

### LINKED ISSUE #104: Reproducible test_dataset crash on Windows / 1.10.2 from anaconda

magnusuMET:
The allocator functions are not protected by a mutex, as the reentrant mutex is defined in `hdf5` which we can't depend on. Is there a way of moving this mutex to `hdf5-sys`? Should we instead move `hdf5-types` to a module within `hdf5`'?

CI results are available at https://github.com/magnusuMET/hdf5-rust/runs/945773746?check_suite_focus=true

Fixes #104

### DIFF
--- a/src/hl/container.rs
+++ b/src/hl/container.rs
@@ -127,11 +127,11 @@ impl<'a> Reader<'a> {
             let mspace = Dataspace::try_new(&out_shape, false)?;
             let size = out_shape.iter().product();
             let mut vec = Vec::with_capacity(size);
+
+            self.read_into_buf(vec.as_mut_ptr(), Some(&fspace), Some(&mspace))?;
             unsafe {
                 vec.set_len(size);
             }
-
-            self.read_into_buf(vec.as_mut_ptr(), Some(&fspace), Some(&mspace))?;
 
             let arr = ArrayD::from_shape_vec(reduced_shape, vec)?;
             Ok(arr.into_dimensionality()?)
@@ -157,10 +157,12 @@ impl<'a> Reader<'a> {
     pub fn read_raw<T: H5Type>(&self) -> Result<Vec<T>> {
         let size = self.obj.space()?.size();
         let mut vec = Vec::with_capacity(size);
-        unsafe {
-            vec.set_len(size);
-        }
-        self.read_into_buf(vec.as_mut_ptr(), None, None).map(|_| vec)
+        self.read_into_buf(vec.as_mut_ptr(), None, None).map(|_| {
+            unsafe {
+                vec.set_len(size);
+            };
+            vec
+        })
     }
 
     /// Reads a dataset/attribute into a 1-dimensional array.
--- a/src/hl/container.rs
+++ b/src/hl/container.rs
@@ -7,6 +7,7 @@ use ndarray::{SliceInfo, SliceOrIndex};
 
 use hdf5_sys::h5a::{H5Aget_space, H5Aget_storage_size, H5Aget_type, H5Aread, H5Awrite};
 use hdf5_sys::h5d::{H5Dget_space, H5Dget_storage_size, H5Dget_type, H5Dread, H5Dwrite};
+use hdf5_sys::h5p::H5Pcreate;
 
 use crate::internal_prelude::*;
 
@@ -44,13 +45,15 @@ impl<'a> Reader<'a> {
         file_dtype.ensure_convertible(&mem_dtype, self.conv)?;
         let (obj_id, tp_id) = (self.obj.id(), mem_dtype.id());
 
-        let fspace_id = fspace.map_or(H5S_ALL, |f| f.id());
-        let mspace_id = mspace.map_or(H5S_ALL, |m| m.id());
-
         if self.obj.is_attr() {
             h5try!(H5Aread(obj_id, tp_id, buf as *mut _));
         } else {
-            h5try!(H5Dread(obj_id, tp_id, mspace_id, fspace_id, H5P_DEFAULT, buf as *mut _));
+            let fspace_id = fspace.map_or(H5S_ALL, |f| f.id());
+            let mspace_id = mspace.map_or(H5S_ALL, |m| m.id());
+            let xfer =
+                PropertyList::from_id(h5call!(H5Pcreate(*crate::globals::H5P_DATASET_XFER))?)?;
+            crate::hl::plist::set_vlen_manager_libc(xfer.id())?;
+            h5try!(H5Dread(obj_id, tp_id, mspace_id, fspace_id, xfer.id(), buf as *mut _));
         }
         Ok(())
     }
@@ -247,12 +250,11 @@ impl<'a> Writer<'a> {
         mem_dtype.ensure_convertible(&file_dtype, self.conv)?;
         let (obj_id, tp_id) = (self.obj.id(), mem_dtype.id());
 
-        let fspace_id = fspace.map_or(H5S_ALL, |f| f.id());
-        let mspace_id = mspace.map_or(H5S_ALL, |m| m.id());
-
         if self.obj.is_attr() {
             h5try!(H5Awrite(obj_id, tp_id, buf as *const _));
         } else {
+            let fspace_id = fspace.map_or(H5S_ALL, |f| f.id());
+            let mspace_id = mspace.map_or(H5S_ALL, |m| m.id());
             h5try!(H5Dwrite(obj_id, tp_id, mspace_id, fspace_id, H5P_DEFAULT, buf as *const _));
         }
         Ok(())
--- a/src/hl/plist.rs
+++ b/src/hl/plist.rs
@@ -1,10 +1,12 @@
 use std::fmt::{self, Debug, Display};
 use std::ops::Deref;
+use std::panic;
 use std::ptr;
 use std::str::FromStr;
 
 use hdf5_sys::h5p::{
     H5Pcopy, H5Pequal, H5Pexist, H5Pget_class, H5Pget_class_name, H5Pget_nprops, H5Piterate,
+    H5Pset_vlen_mem_manager,
 };
 
 use crate::internal_prelude::*;
@@ -200,6 +202,26 @@ impl PropertyList {
             PropertyListClass::from_str(&name)
         })
     }
+}
+
+/// Set the memory manager for variable length items to
+/// the same allocator as is in use by hdf5-types
+// TODO: move this to dataset_transfer module when DatasetTransfer plist is implemented
+pub fn set_vlen_manager_libc(plist: hid_t) -> Result<()> {
+    extern "C" fn alloc(size: size_t, _info: *mut c_void) -> *mut c_void {
+        panic::catch_unwind(|| unsafe { libc::malloc(size) }).unwrap_or(ptr::null_mut())
+    }
+    extern "C" fn free(ptr: *mut c_void, _info: *mut libc::c_void) {
+        let _ = panic::catch_unwind(|| unsafe { libc::free(ptr) });
+    }
+    h5try!(H5Pset_vlen_mem_manager(
+        plist,
+        Some(alloc),
+        ptr::null_mut(),
+        Some(free),
+        ptr::null_mut()
+    ));
+    Ok(())
 }
 
 #[cfg(test)]
--- a/hdf5-types/src/lib.rs
+++ b/hdf5-types/src/lib.rs
@@ -1,5 +1,4 @@
 #![recursion_limit = "1024"]
-#![cfg_attr(feature = "cargo-clippy", allow(clippy::transmute_bytes_to_str))]
 #![cfg_attr(feature = "cargo-clippy", allow(clippy::missing_safety_doc))]
 
 #[cfg(test)]
--- a/hdf5-types/src/string.rs
+++ b/hdf5-types/src/string.rs
@@ -245,7 +245,7 @@ impl VarLenAscii {
 
     #[inline]
     pub fn as_str(&self) -> &str {
-        unsafe { mem::transmute(self.as_bytes()) }
+        unsafe { str::from_utf8_unchecked(self.as_bytes()) }
     }
 
     #[inline]
@@ -439,7 +439,7 @@ impl<A: Array<Item = u8>> FixedAscii<A> {
 
     #[inline]
     pub fn as_str(&self) -> &str {
-        unsafe { mem::transmute(self.as_bytes()) }
+        unsafe { str::from_utf8_unchecked(self.as_bytes()) }
     }
 
     #[inline]
@@ -549,7 +549,7 @@ impl<A: Array<Item = u8>> FixedUnicode<A> {
 
     #[inline]
     pub fn as_str(&self) -> &str {
-        unsafe { mem::transmute(self.as_bytes()) }
+        unsafe { str::from_utf8_unchecked(self.as_bytes()) }
     }
 
     #[inline]


aldanor:
This is pretty sad :( I really wanted to have hdf5-types as a standalone thing because in reality (varlen allocator aside) it doesn't have anything to do with libhdf5.

Here's another thing (something I was thinking about for a while):
- For *writing*, we don't really care what allocator is there etc. HDF5 simply takes your pointers and reads them. Hence it doesn't matter whether it's libc-allocated or whatever else. So, for writing, hdf5-types is perfectly fine.
- Even more: for writing, the data doesn't have to be owned. I have considered for a while that maybe there should be a `FixedStringView` and `VarLenStringView`, etc. This would also allow to interface better with rust strings without huge waste of memory.
- For *reading*, hdf5-types fixed types are fine. It's only for reading varlen arrays is where this kicks in and we have to use a matching deallocator. Wonder if there's some cleaner way to do it than making the entire hdf5-types depend on hdf5-sys? Hm..

aldanor:
From the docs:

> Even when using this function, it is best where possible to ensure that all components of a C application are built with the same version of Visual Studio and configuration (Debug or Release), and thus linked against the same CRT.

And also:

> Use this function only to allocate memory inside third-party HDF5 filters. It will generally not be safe to use this function to allocate memory for any other purpose.

Would it be possible, as they say, to "ensure that all components are built with the same version" on Windows maybe? This might prevent other bugs as well, and will get rid of this varlen allocator problem?

aldanor:
As for the mutex, maybe it should live in a separate crate? Like `hdf5-lock`. Would there be any extra need for that?

magnusuMET:
Maybe it's possible to make two variants of the varlen structs, where one permits reading (implements `Drop` using `H5free_memory`), and the other is a `StringView<'a>`/`ArrayView<'a>` which only borrows an item for writing to a dataset?

magnusuMET:
> As for the mutex, maybe it should live in a separate crate? Like `hdf5-lock`. Would there be any extra need for that?

It would be nice to have such a lock in `hdf5-sys`, or make it depend on `hdf5-sys` so it will always resolve to the same mutex (`hdf5-sys` has the `links` attribute). I would use such a mutex in the `netCDF` crate to allow simultaneous use of `hdf5` and `netCDF`.

magnusuMET:
This version should be safe to use, apart from a use-after-free if calling `H5close`, which would clean up all memory used by the library. This should be covered as `H5close` is not being used in the higher level bindings.

magnusuMET:
I see the allocation mechanisms are not available in the older `hdf5` versions. What is the best way to propagate this information? Do we have to use another build script?

aldanor:
@magnusuMET As you can see, it is a much wider issue than just changing the impl of `Drop` :) It's not a problem to implement all this, more of figuring things out and agreeing on what makes most sense.

1. First and foremost, `hdf5-types` now depends on `hdf5-sys`... This is a bit unfortunate, as types like `FixedString`, `FixedArray` etc have nothing to do with libhdf5 whatsoever; you could in theory use `hdf5-types` as a dependency without having to depend on hdf5 at all, but I'm not sure anyone would be actually doing that? If `hdf5-types` is no longer detached, I'm not sure if it even makes sense for it to remain a separate crate...
2. VarLen objects constructor and destructor may now crash if `hdf5` is not initialized, even if you use it outside of hdf5 context entirely. I'm not sure how the allocator mechanic works with `H5open()` and `H5close()`, and nothing in `hdf5-sys`/`hdf5-types` is actually forcing the library to initialize. Needs testing (i.e. what happens if you create a varlen string before H5open was called, or after the library was closed).
3. VarLen objects constructor and destructor are now locked behind a mutex. Should not matter much generally.
4. `H5allocate_memory` and `H5deallocate_memory` are only available in 1.18.15 and later. What to do in earlier versions? Given that 1.8.15 is over 5 years old now, we may, of course, just require it as a minimum version and disallow all earlier versions strictly (technically that's yet another breaking change, but it might make sense in the long run). This would also allow to not have to write another build script and propagate version info.

// On a somewhat related note, I was thinking if it was possible to integrate support for `String`, `&str`, `Vec<T>` and `&[T]`, even if the user has to pay the price of partial reallocation (chunk by chunk, say) – seeing how many people have problems with varlen/fixed arrays/strings, and given that there's no built-in conversion paths between fixed and varlen strings in HDF5 for some reason. Unfortunately `String` is laid out as `(ptr, capacity, allocator, len)` and HDF5 accepts `(ptr, len)` – but in theory you could take user's array of data with rust strings, mutate it temporarily, swapping len and capacity, and then mutate it back. It's quite evil and breaks a few Rust paradigms of course :)

magnusuMET:
Regarding 2., reading the source (1.10.6), `H5allocate_memory` will resolve to `malloc`/`calloc`, and `H5free_memory` will resolve to `free`, expect when using a debug flag to check for memory sanity. Not sure whether any releases actually use that flag. We can emit a warning/error if compiling an older version than supported if using a DLL version to warn that a different `crt` might be in use. In this case we could use the system allocator to get memory and leak memory returned from functions.

> but in theory you could take user's array of data with rust strings, mutate it temporarily, swapping len and capacity, and then mutate it back.

Would love to see a prototype of this :evil:

aldanor:
@magnusuMET Hm! Check [this](https://support.hdfgroup.org/HDF5/doc/RM/H5P/H5Pset_vlen_mem_manager.htm) out:

> H5Pset_vlen_mem_manager sets the memory manager for variable-length datatype allocation in H5Dread and free in H5Dvlen_reclaim.

> The alloc and free parameters identify the memory management routines to be used. If the user has defined custom memory management routines, alloc and/or free should be set to make those routine calls (i.e., the name of the routine is used as the value of the parameter); if the user prefers to use the system's malloc and/or free, the alloc and free parameters, respectively, should be set to NULL

Isn't that like exactly what we want?

Need to research this further and maybe read the sources too.

magnusuMET:
@aldanor Great find! That is exactly the function we need! I've added a function which uses the system allocator, using the CRT of the caller.

I am not sure what the plist should be, or where it would be the best to set this. The source `src/H5Pdxpl.c` says this is allowed on `H5P_DATASET_XFER` objects.

In the future one could implement this using `std::alloc`, but this requires tracking the allocations  through e.g. `free_info` and `alloc_info` to preserve the layouts of the items.

aldanor:
The DXPL plists are 'transfer plists', they are not attached to anything and are assigned per operation (https://support.hdfgroup.org/HDF5/doc/RM/RM_H5D.html#Dataset-Read).

So we could just ensure to attach a proper dxpl plist to each `H5Dread()` and that's it, `libc::free()` guaranteed for all read operations (initiated by `hdf5` crate), so if it works out there's no need for any changes in `hdf5-types` I think.

magnusuMET:
@aldanor No changes are needed in `hdf5-types`!

I've added another trait function to `H5Type` so we only need to create this property list when dealing with vlen types. I do struggle with the macros that implements this type for tuples and `derive`. Is this too much premature optimisation and should we just attach the plist for all read/writes?

aldanor:
With writes we don't care I think? (But can still attach it)

For reads, what's the downside to attaching it always? I.e., why not just always attach it on all reads? Also, it's quite likely system allocator would be more efficient then their home-baked one.

aldanor:
Ok, just read your code – yes, technically you could do it like this, I guess, but it's a bit more complicated. It's more like `has_varlen()`, because of compound types, arrays and all other types of composition (fixed array of compound types where one field is varlen), etc.

There's actually an H5T method for this – [`H5Tdetect_class()`](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5T.html#Datatype-DetectClass), you could probably just call that at reading/writing time instead of messing with TypeDescriptor I guess. That's option 1.

Option 2 is always using it; why not? It would not affect non-varlen stuff anyways (I *think*). Not sure if it would affect conversion buffers etc, but probably not.

magnusuMET:
@aldanor I was considering not using it to avoid two function calls, but these aren't particularly expensive. Should be ready to merge now. CI is green for the `feature/gha` branch with this merged.

I added the `catch_unwind` in case a future implementation wishes to implement alloc/free using the rust allocator. Such panic catchers should be added to all the functions we use as function pointers.

src/hl/plist.rs
```diff
@@ -202,6 +203,20 @@ impl PropertyList {
     }
 }
 
+pub fn set_vlen_manager(plist: crate::internal_prelude::hid_t) -> Result<()> {
```
aldanor, line 206:
```suggestion
pub fn set_vlen_manager(plist: hid_t) -> Result<()> {
```

src/hl/plist.rs
```diff
@@ -202,6 +203,20 @@ impl PropertyList {
...
+    use std::panic;
+    extern "C" fn alloc(size: libc::size_t, _info: *mut libc::c_void) -> *mut libc::c_void {
+        panic::catch_unwind(|| unsafe { libc::malloc(size) }).unwrap_or(ptr::null_mut())
+    }
+    extern "C" fn free(ptr: *mut libc::c_void, _info: *mut libc::c_void) {
+        let _ = panic::catch_unwind(|| unsafe { libc::free(ptr) });
+    }
+    h5try!(unsafe {
+        H5Pset_vlen_mem_manager(plist, Some(alloc), ptr::null_mut(), Some(free), ptr::null_mut())
+    });
```
aldanor, line 216:
```suggestion
    h5try!(
        H5Pset_vlen_mem_manager(plist, Some(alloc), ptr::null_mut(), Some(free), ptr::null_mut())
    );
```

src/hl/plist.rs
```diff
@@ -202,6 +203,20 @@ impl PropertyList {
     }
 }
 
+pub fn set_vlen_manager(plist: crate::internal_prelude::hid_t) -> Result<()> {
+    use std::panic;
+    extern "C" fn alloc(size: libc::size_t, _info: *mut libc::c_void) -> *mut libc::c_void {
```
aldanor, line 208:
```suggestion
    extern "C" fn alloc(size: size_t, _info: *mut c_void) -> *mut c_void {
```

src/hl/plist.rs
```diff
@@ -202,6 +203,20 @@ impl PropertyList {
...
+        panic::catch_unwind(|| unsafe { libc::malloc(size) }).unwrap_or(ptr::null_mut())
+    }
+    extern "C" fn free(ptr: *mut libc::c_void, _info: *mut libc::c_void) {
```
aldanor, line 211:
```suggestion
    extern "C" fn free(ptr: *mut c_void, _info: *mut c_void) {
```

src/hl/plist.rs
```diff
@@ -202,6 +203,20 @@ impl PropertyList {
     }
 }
 
+pub fn set_vlen_manager(plist: crate::internal_prelude::hid_t) -> Result<()> {
+    use std::panic;
```
aldanor, line 207: Can this import be moved all the way to the top, with other std imports? (to match the style)

src/hl/plist.rs
aldanor, line 206: Maybe add `// TODO: move this to dataset_transfer module when DatasetTransfer plist is implemented`? (so it doesn't hang out here in the wild)

Also, I don't think this name is clear enough; two alternative options: 

1. `pub fn set_vlen_manager_libc(plist: hid_t)`
2. `pub fn set_vlen_manager<A: Fn(usize) -> *mut u8, F: Fn(*mut u8)>(plist: hid_t, alloc: A, free: F)`

In (2), you'd call it as

```rust
set_vlen_manager(plist_id, |size| unsafe { libc::malloc(size) }, |ptr| unsafe { libc::free(ptr) })
```

(assuming the catch_unwind stuff can remain in the generic function itself)

You could probably even have both, where (1) calls (2).

src/hl/container.rs
```diff
@@ -44,13 +44,16 @@ impl<'a> Reader<'a> {
...
-        let fspace_id = fspace.map_or(H5S_ALL, |f| f.id());
-        let mspace_id = mspace.map_or(H5S_ALL, |m| m.id());
-
         if self.obj.is_attr() {
             h5try!(H5Aread(obj_id, tp_id, buf as *mut _));
         } else {
-            h5try!(H5Dread(obj_id, tp_id, mspace_id, fspace_id, H5P_DEFAULT, buf as *mut _));
+            let fspace_id = fspace.map_or(H5S_ALL, |f| f.id());
+            let mspace_id = mspace.map_or(H5S_ALL, |m| m.id());
+            let xfer = PropertyList::from_id(h5call!(hdf5_sys::h5p::H5Pcreate(
```
aldanor, line 52: `use hdf5::sys::h5p::H5Pcreate` at the top?

src/hl/container.rs
```diff
@@ -157,10 +160,12 @@ impl<'a> Reader<'a> {
...
-        unsafe {
-            vec.set_len(size);
-        }
-        self.read_into_buf(vec.as_mut_ptr(), None, None).map(|_| vec)
+        self.read_into_buf(vec.as_mut_ptr(), None, None).map(|_| {
+            unsafe {
+                vec.set_len(size);
+            };
+            vec
+        })
```
aldanor, line 168: 👍

aldanor:
@magnusuMET I've added a few comments here and there, but looks good.

Re: catch_unwind, maybe add a separate ticket so we don't forget to review where else it might be needed?

src/hl/plist.rs
magnusuMET: > Maybe add `// TODO: move this to dataset_transfer module when DatasetTransfer plist is implemented`? (so it doesn't hang out here in the wild)
> 
> Also, I don't think this name is clear enough; two alternative options:
> 
>     1. `pub fn set_vlen_manager_libc(plist: hid_t)`
> 
>     2. `pub fn set_vlen_manager<A: Fn(usize) -> *mut u8, F: Fn(*mut u8)>(plist: hid_t, alloc: A, free: F)`
> 
> 
> In (2), you'd call it as
> 
> ```rust
> set_vlen_manager(plist_id, |size| unsafe { libc::malloc(size) }, |ptr| unsafe { libc::free(ptr) })
> ```
> 
> (assuming the catch_unwind stuff can remain in the generic function itself)
> 
> You could probably even have both, where (1) calls (2).

(2) requires quite a bit of trickery, as we have to get a function pointer from a closure. The easiest way would be to get a closure pointer, wrap this up in the `alloc_info` and `free_info`, and call this from a trampoline. This is a bit overkill when we are tied to the same allocator as for the varlen types in `h5-types` .

magnusuMET: (2) requires quite a bit of trickery, as we have to get a function pointer from a closure. The easiest way would be to get a closure pointer, wrap this up in the alloc_info and free_info, and call this from a trampoline. This is a bit overkill when we are tied to the same allocator as for the varlen types in h5-types .

magnusuMET:
@aldanor Thanks for the review! Hopefully ready to merge now.

Opened #110 for tracking the panic safety

src/hl/plist.rs
aldanor: Re: (2), yea, that's kind of what I was thinking, that's why the info is there :) (There's a similar example in Rustonomicon, I think)

The reason I mentioned it - we'll get to adding proper dxpl plists soon (there's just a few functions) - there we'll need to implement this in the most generic fashion.

For now it's ok I guess as long as there's a todo note.

magnusuMET: I have also done something similar in a c-wrapper for [`uvc`](https://github.com/mulimoen/libuvc-rs/blob/3394f0b7aecd315a23e0761f1cbd4ce713a413ff/src/streaming.rs#L51). It's a bit tricky to keep the lifetimes in order!

aldanor:
Good to merge and one less bug, thanks for taking time to look at this :)

(re: DXPL and all that, I'll do it myself a bit later)

PR #117, opened 2020-08-19
Add deprecated functions and deprecation warnings

### LINKED ISSUE #114: Alias the latest API available per hdf5 version
`hdf5-c` introduces new functions with labels `func1`, `func2`, and selects the correct version based on a compile time macro.

We should to go through `hdf5-sys` and alias the latest version available to the common name `func`, as well as always being available through `func1` (with deprecation warning where applicable) and `func2` (where hdf5 version allows).

See #95 for more information

mulimoen:
I went through the [documentation](https://portal.hdfgroup.org/display/HDF5/HDF5) and added all functions which were marked as deprecated, mostly for completeness of the bindings. Most of these functions were deprecated in 1.8.0, which is a version we don't support.

 I've also fixed up the name of some structs to be consistent with the naming in 1.12.0, and made sure we have aliases for these to remain compatible with other supported versions.

Comments on objects are officially deprecated in favor of attributes. I've added the `deprecated` attribute on the set/clear_comment in the top crate.

FIxes #114

### DIFF
--- a/hdf5-sys/src/h5a.rs
+++ b/hdf5-sys/src/h5a.rs
@@ -23,7 +23,7 @@ impl Default for H5A_info_t {
     }
 }
 
-#[deprecated(note = "Deprecated in HDF5 1.8.0, use H5A_operator2_t")]
+#[deprecated(note = "deprecated in HDF5 1.8.0, use H5A_operator2_t")]
 pub type H5A_operator1_t = Option<
     extern "C" fn(
         location_id: hid_t,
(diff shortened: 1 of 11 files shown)


hdf5-sys/src/h5a.rs
```diff
@@ -92,4 +105,19 @@ extern "C" {
     pub fn H5Aexists_by_name(
         obj_id: hid_t, obj_name: *const c_char, attr_name: *const c_char, lapl_id: hid_t,
     ) -> htri_t;
+
+    #[deprecated(note = "Deprecated in HDF5 1.8.0, use H5Aget_info()")]
```
aldanor, line 109: Use consistent casing everywhere? I.e. `note = "deprecated ..."`

hdf5-sys/src/h5a.rs
```diff
@@ -92,4 +105,19 @@ extern "C" {
...
+    pub fn H5Aget_num_attrs(loc_id: hid_t) -> c_int;
+    #[deprecated(note = "Deprecated in HDF5 1.8.0, use H5Aopen_by_idx()")]
```
aldanor, line 111: Either use "()" everywhere, or drop it everywhere?

aldanor:
Are we planning to go ahead with this? Will require a rebase if so.

mulimoen:
I would like to see this included, it makes handling different versions of `hdf5` a bit easier, and more obvious when one is using these deprecated functions.

aldanor:
Almost good to go!

mulimoen:
Have you left some review comments I can't see?

aldanor:
> Have you left some review comments I can't see?

Sorry I'm dumb, opened a pending a review and haven't submitted it :) There you go ^

aldanor (approved):

PR #121, opened 2020-09-27
Support static cross-compilation from Unix to Windows.

hugwijst:
Use wine to run HDF5 configuration programs.

To run Windows unit tests on Linux, use wine as the runner. For example:
```
CARGO_TARGET_X86_64_PC_WINDOWS_GNU_RUNNER=wine64 cargo test --features hdf5-sys/static --target x86_64-pc-windows-gnu
```

If your main wine configuration is 32-bit but you're cross-compiling for 64-bit windows, you will need to specify a different wine environment using the `WINEPREFIX` environment variable, eg. `WINEPREFIX=~/wine_win64`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,3 +205,18 @@ jobs:
           msiexec /i ${{env.MSI_PATH}} /quiet /qn /norestart
       - name: Build and test all crates
         run: cargo test -vv
+
+  wine:
+    name: wine
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with: {submodules: true}
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with: {toolchain: stable, target: x86_64-pc-windows-gnu, profile: minimal, override: true}
+      - name: Install dependencies
+        run: sudo apt-get install wine64 mingw-w64
+      - name: Build and test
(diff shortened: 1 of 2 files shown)


magnusuMET:
You might want to add another check to the CI to ensure we don't accidentally break this. Something like the following should do the trick:
```yml
  wine:
    name: wine
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@v2
        with: {submodules: true}
      - name: Install Rust
        uses: actions-rs/toolchaing@v1
        with: {toolchain: stable, profile: minimal, override: true}
      - name: Install dependencies
        run: sudo apt-get install wine64 libhdf5-dev
      - name: Build and test
        run: env CARGO_TARGET_X86_64_PC_WINDOWS_GNU_RUNNER=wine64 cargo test --features hdf5-sys/static --target x86_64-pc-windows-gnu
```

jmrgibson:
@magnusuMET  anything else needed for this to go in? This would be handy for us too.

magnusuMET:
Only @aldanor can merge and publish this

mulimoen:
Did a rebase to run this through CI

aldanor:
Not familiar with these windows-specific details, but I guess if it builds green and everyone's happy - should be good to go.

mulimoen:
Merging this, thanks @hugwijst for making cross-compilation easier!

PR #129, opened 2021-01-26
Fix CI

mulimoen:
This fixes various CI failures, and adds a build for windows 1.12

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,8 @@ name: CI
 
 on:
   push:
+    branches:
+      - master
   pull_request:
     branches:
       - master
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -124,6 +126,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
         with: {submodules: true}
+      - name: Change to older toolchain
+        if: matrix.os == 'macos'
+        run: sudo xcode-select -s "/Applications/Xcode_11.7.app"
```
aldanor, line 131: @mulimoen Just wondering why is this needed?

aldanor (approved):
Looks great, thanks!

.github/workflows/ci.yml
mulimoen: The error message: `implicit declaration of function 'H5O__fsinfo_set_version' is invalid in C99 `.

The updated compiler is a bit better at enforcing `c99` standards, and `hdf5` placed the function definition below first use. This should be fixed in updated version of `hdf5`, we should update `hdf5-src` to pull the official fix.

mulimoen: See #130 for how to fix this permanently

Issue #122, opened 2020-11-23
How to use compression properly


ogiermaitre:
This is most certainly a sill question, but I'm not able to use compression properly.
This is the code I used: 
``` rust

let mut tmp = Vec::new();

// [...] here I put stuff into the vector

let mut ds_builder = group.new_dataset::<YearlyEntry>();
let ds_builder = ds_builder.gzip(4);

let ds = ds_builder.create(project, (tmp.len(), 1)).unwrap();
ds.write_raw(&tmp).unwrap();


```

At the end, the file is pretty large (~2.1GB), and I'm able to zip it to a ~90MB file. Did I miss something ?

magnusuMET:
@ogiermaitre You can check if the variable has the `gzip` filter using `h5stat`. If it is set you might have to adjust the `chunk`-size as the automatic chunk size might be a poor choice

ogiermaitre:
Thanks for your fast answer. 
The compression seems to be activated in h5dump, for my dataset:
```
Dataset layout information:
	Dataset layout counts[COMPACT]: 0
	Dataset layout counts[CONTIG]: 0
	Dataset layout counts[CHUNKED]: 1
	Dataset layout counts[VIRTUAL]: 0
	Number of external files : 0
Dataset filters information:
	Number of datasets with:
		NO filter: 0
		GZIP filter: 1
		SHUFFLE filter: 0
		FLETCHER32 filter: 0
		SZIP filter: 0
		NBIT filter: 0
		SCALEOFFSET filter: 0
		USER-DEFINED filter: 0
```

I tried to set the chunk_size, but I didn't find the right function to do it, so I tried the DatasetBuilder::chunk method, without real improvement (it's not really clear to me, how I can compute the chunk shape, in order to set a given chunk_size).

No compression: 2.2 GB
GZIP standard shape: 2.1GB (num_chunks: 1024).
GZIP with shape (1_000, 1): 2.1GB (num_chunks: 1505)
GZIP with shape (1_000_000,1): 2.1GB (num_chunks 2).

Is there a way to set the chunk_size directly ?
I can provide the complete h5stat if it helps.

magnusuMET:
The `chunk` method on the `DatasetBuilder` should set the `chunk_size`. Is `YearlyEntry` a numeric type or some custom type?

ogiermaitre:
Is something like: `ds_builder.chunk((10_000,1));` a valide configuration ?

The type is indeed a custom type, and is defined like that: 

[...]

PR #131, opened 2021-01-27
use const generics in hdf5-types for array impl

mulimoen:
Const generics are rapidly approaching which means `impl_array` can be replaced with a generic size, supporting more array cases than one can manually specify. This is feature gated on `const_generics` on `hdf5-types` and should only be a minor bump.

Ref #122

### DIFF
--- a/hdf5-types/src/array.rs
+++ b/hdf5-types/src/array.rs
@@ -14,49 +14,78 @@ pub unsafe trait Array: 'static {
     fn capacity() -> usize;
 }
 
-macro_rules! impl_array {
-    () => ();
+#[cfg(not(feature = "const_generics"))]
+mod impl_array {
+    use super::*;
 
-    ($n:expr, $($ns:expr,)*) => (
-        unsafe impl<T: 'static> Array for [T; $n] {
-            type Item = T;
+    macro_rules! impl_array {
+        () => ();
 
-            #[inline(always)]
-            fn as_ptr(&self) -> *const T {
-                self as *const _ as *const _
+        ($n:expr, $($ns:expr,)*) => (
+            unsafe impl<T: 'static> Array for [T; $n] {
+                type Item = T;
+
+                #[inline(always)]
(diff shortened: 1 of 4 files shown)


aldanor:
I wonder, if we can enable this feature automatically? (e.g. through build script, hmm) There would be no reason not to use it once 1.51 is out.

Also, need some checks that it compiles - we can either employ nightly for that, or just wait till 1.51 and add checks on stable.

mulimoen:
We could enable this using `autocfg` but we will need another `build.rs`... I think it is ok to have a feature for this for now. We might want to make this `default` when 1.51 is out and remove the feature entirely when enough time has passed.

hdf5-types/src/array.rs
```diff
@@ -14,49 +14,76 @@ pub unsafe trait Array: 'static {
     fn capacity() -> usize;
 }
 
-macro_rules! impl_array {
-    () => ();
+#[cfg(not(feature = "const_generics"))]
+mod impl_array {
+    use super::*;
```
aldanor, line 19: Minor - missing newline after `use`

hdf5-types/src/array.rs
```diff
@@ -14,49 +14,76 @@ pub unsafe trait Array: 'static {
...
+    );
+}
 
-            #[inline(always)]
-            fn as_mut_ptr(&mut self) -> *mut T {
-                self as *mut _ as *mut _
-            }
+#[cfg(feature = "const_generics")]
+mod impl_array {
+    use super::*;
```
aldanor, line 67: Same, minor - missing newline after `use`

aldanor:
Very minor comment aside, looks great and lgtm!

aldanor (approved):

mulimoen:
@aldanor Maybe it is worthwhile to publish this to `crates.io`?

aldanor:
I wonder if we should've added `const-generics` feature to the root crate, hmm...

PR #134, opened 2021-01-27
0.7.1

aldanor:


aldanor:
@mulimoen all 0.7.1 versions should be on crates.io now (and in docs.rs queue); I've added a GH release as well.

PR #138, opened 2021-02-03
simple optional path to download HDF5 from conda and statically link

pmarks:
We've been maintaining this as separate branch for a while: it works really well for us to get reliable builds on diverse machines without requiring an HDF5 installation & to have unit test in downstream crates work seamlessly.  

Is this something you'd be open to? Any suggestions on how you'd like to approach it differently?

### DIFF
--- a/hdf5-sys/build.rs
+++ b/hdf5-sys/build.rs
@@ -640,6 +640,13 @@ impl Config {
 }
 
 fn main() {
+
+    #[cfg(feature = "conda")]
+    if env::var("CARGO_FEATURE_CONDA").is_ok() {
+        conda_dl::conda_static();
+        return
+    }
+
     if feature_enabled("STATIC") && std::env::var_os("HDF5_DIR").is_none() {
         get_build_and_emit();
     } else {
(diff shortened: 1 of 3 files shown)


hdf5-sys/build.rs
```diff
@@ -683,3 +690,152 @@ fn get_build_and_emit() {
...
+                    panic!("check sum of downloaded archive is incorrect: md5sum={}", sum);
+                }
+            }
+        }
+    
+        println!("cargo:rustc-link-search={}", out_dir.join(conda::LIB_PATH).display());
+    
+        #[cfg(target_os = "windows")]
+        {
+            println!("cargo:rustc-link-lib=static=zlibstatic");
```
mulimoen, line 822: Will this clash with `zlib-sys`?

mulimoen:
This is a whole lot of (optional) dependencies to add to the project to solve something package managers should be able to do. We also have the `static` build for easy bundling, although this adds a lot of build-time. Would a solution be to adapt the build script to allow `HDF5_DIR` to accept a statically built library?

If we go ahead with a pure conda option we should add a new crate (`hdf5-conda`?) and instead plumb it like `hdf5-src`.

mulimoen:
I've had a go at this which resulted in [the following branch](https://github.com/mulimoen/hdf5-rust/tree/feature/conda). This moves the extraction to a helper crate, decreasing dev-dependencies. It seems dependencies for hdf5-c must also be downloaded and linked appropriately (on non-linux), which might conflict badly with libraries installed from elsewhere. Is support for macos and windows necessary, or could somebody with experience maybe try to fix this for these platforms?

PR #139, opened 2021-02-19
Revamp object identifier ownership

### LINKED ISSUE #76: possible memory leak
Hi, I've been working on a project that goes through all chunks of all datasets in a file. And then drops all references to the original hdf5::File. It seems that this causes a memory leak, but I'm not 100% sure. The structure I've built up is just a couple of MBs or less for a 3.7gb input file (or 8mb serialized to json), but after having built the index I am using about 700 mb of ram (vs just a few mbs if I only open and deserialize the same index). I have no idea if the leak is in hdf5-rust or perhaps more probable in native hdf5 itself. Relevant code: https://github.com/gauteh/dars/blob/hdfidx/hidefix/src/idx/dataset.rs#L86

### LINKED ISSUE #137: Memory leak in group creation
The following is a reproducer for what seems like a fatal memory leak.

Rust crate version: 0.7.1
HDF5 Version: 1.12.0
OS: MacOS, Linux

**Memory usage grows linearly from 10MB to 2GB over 100,000 iterations**
```Rust
fn write_test_file(fname: &str, iteration: usize, num_groups: usize) -> Result<(), Box<dyn std::error::Error>> {
    println!("iteration {}", iteration);
    let h5f = hdf5::File::create(fname)?;
    for j in 0..num_groups {
        h5f.create_group(&format!("group-{}", j))?;
    }
    Ok(())
}

fn main() -> Result<(), Box<dyn std::error::Error>> {
    for i in 0..100000 {
        write_test_file("test.h5", i, 100)?
    }
    Ok(())
}
```

**Memory usage is stable at 15MB**
```C
#include <stdlib.h>
#include <hdf5.h>

void write_test_file(const char* fname, int iteration, int array_size, int num_groups)
{
    printf("iteration %d\n", iteration);
    hid_t h5f = H5Fcreate(fname, H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);

    for (int j = 0; j < num_groups; ++j)
    {
        char gname[1024];
        snprintf(gname, 1024, "group-%d", j);
        hsize_t dims[3] = {array_size, array_size, array_size};
        hid_t h5g = H5Gcreate(h5f, gname, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
        H5Gclose(h5g);
    }
    H5Fclose(h5f);
}

int main()
{
    for (int i = 0; i < 100000; ++i)
    {
        write_test_file("test.h5", i, 10, 100);
    }
    return 0;
}
```

mulimoen:
This is a new way of handling object identifiers given by us from the `hdf5-c` library. 

The old approach was to store the handles in a registry, a hashmap that kept track of all previous identfiers. Each object addtionally kept track of whether it is invalid or not through an `RwLock`. This works very well for our tests, but fails when opening and closing many objects (#76, #137).

This PR instead relies on `hdf5-c` to keep track of the identifiers and validity. The `Handle` is now a type which owns the underlying object directly, and ownership is a bit stricter. 

In practical terms this does not change much. Opened file objects will not be automatically closed when a file is dropped, the user must set the strong close degree on the file to get this behaviour. Setting the strong close degree is dangerous (in the sense it closes the wrong items), but not unsafe, and a warning has been added to the corresponding function.

The idea is to strengthen the relationship between an object handle on the `rust` side and the `hdf5` side. A handle owns the object, and is cloned by incrementing the reference count and handing out a new object. The old approach was to try and close all related handles, acting as a half `FileCloseDegree::Strong`. The new approach requires the user to explicitely opt into this behaviour, with all the pitfalls this entails...

Soundness difficulties (but not memory safety concern) leading to double free:
* Creating a new borrowed object from an ID without incref
* `decref` without incref
* The strong file close degree

Should we remove or change some items from our API to remove some of these? We could add a counterpart of `from_id` (`from_borrowed_id`), make `decref` hidden, and use `unsafe` to annotate these functions.

Fixes #76
Fixes #137


mulimoen:
@aldanor It would be nice if you had a look at this proposal. This is the opposite of keeping the registry, but seems to be easier to reason about regarding ownership of objects.

aldanor:
This will need to be rebased as #140 is merged. @mulimoen I've read through this, and I guess my main question is: "why does it work?" :) Especially pre-1.10, for 1.8.x. (specifically, in multi-threaded environment)

Also, for the crate user who doesn't want to dabble in source code of this crate, what would be the major usage differences? (in regards to strong close etc)

mulimoen:
It is simply having a single owner for every ID, taking special care to incref if we share the ID, and never decref more than necessary (special care is taken in tests to avoid this). The previous code was buggy as it always decreffed all open handles on file close. There are no multi-threaded concerns as we always operate on the ID using a mutex (serial execution).

For users this means never use strong close (unless dead certain no objects are open), don't decref (`Drop` is ok), take care during FFI/`hdf5` to never give ownership of an ID without an incref.

mulimoen:
Seems `DynValue` leaks memory, which is picked up by the address sanitizer. I'm debugging this ATM.

mulimoen:
6b66419 changed the inner type of `OwnedDynValue` to a `Box`. This was for making it obvious that the type is fixed-size. If https://github.com/rust-lang/rust/issues/63291 is stabilized we might want to use `new_uninit_slice`.

src/hl/plist/dataset_create.rs
```diff
@@ -741,11 +741,13 @@ impl DatasetCreate {
...
+            let mut out: mem::MaybeUninit<T> = mem::MaybeUninit::uninit();
             let buf = value.get_buf();
-            ptr::copy_nonoverlapping(buf.as_ptr(), (&mut out as *mut T).cast(), buf.len());
-            mem::forget(value);
-            out
+            ptr::copy_nonoverlapping(buf.as_ptr(), out.as_mut_ptr().cast(), buf.len());
+            // Drop the Box<[u8]>, but not subfields
+            value.drop_nonrecursive();
+            // We now have exclusive access to all subfields
+            out.assume_init()
```
aldanor, line 750: Since `drop_nonrecursive()` is only used here, I wonder if it could be made `pub(crate)` and this whole block could be pulled out into a method on `OwnedDynValue` like this:

```rust
    #[doc(hidden)]
   pub unsafe fn cast<T: H5Type>(mut self) -> Result<T> {
        ...
   }
```

where it could also check that `T::type_descriptor() == self.tp` (or it could return `-> T` without `Result` and just panic on mismatch).

And then here it just becomes

```rust
let dtype = Datatype::from_type::<T>()?;
self.get_fill_value(&dtype.to_descriptor()?)?.map(|v| unsafe { v.cast() }).transpose()
```

mulimoen: The new code pushes all the unsafety on `hdf5-types`, making the code a lot clearer. As zero-sized fixed arrays are not possible to create in `hdf5`, I've added a small expection to `dyn_drop` of `DynArray` to support this code path.

aldanor:
Posted the last batch of minor comments ^

I was wondering, do we even really want `incref()` and `decref()` methods on `Handle`? What if we removed both and replaced them with `try_borrow()`? (same as in `Object`)

aldanor:
> I suppose advanced users of `hdf5` want to use underlying functions which may take ownership, requiring `incref`. `decref` may be dangerous and could invalidate a `Handle`, but I guess there are uses?

I think the point is, if you use the high-level interface (i.e. `hdf5`) crate, I can't really imagine a case where you will be increfing or decrefing things (i.e., where `try_new`, `try_borrow`, `clone` and `drop` will not be sufficient for more advanced cases).

If you really need to dig into some advanced territory (i.e. using `hdf5-sys` directly), chances are you will probably be using straight `hid_t` anyway, no?

I'm just trying to think of cases and I can't really find none.

Well, given that this update is already quite big, let's leave it be, but perhaps remove it in a later update (I'll add a ticket) and maybe remove `Handle` altogether then which will simplify things. Now that there's no registry and it's a simple wrapper around an int, I'm not sure it's not needed at all.

aldanor:
Ok well, let's try to merge this (hope it doesn't shoot us in the foot, in some multi-threaded cases or some other weird stuff like that) :)

PR #140, opened 2021-02-25
The DCPL branch

### LINKED ISSUE #45: Reading slices fails with negative indices
Say you have a simple one-dimensional dataset with shape `[64]`. If you try to read the last entry with `dataset.read_slice_1d(s![-1])` hdf5 will error out (of course, the same happens with any dimensional dataset). A current workaround is to set `let shape = dataset.shape();` and then execute `dataset.read_slice_1d(s![shape[0]-1])`.

I will come up with a minimum working example, but just wanted to note this here already, in case somebody encounters the same issue.

### LINKED ISSUE #88: Can't create resizable datasets for only some indices

### LINKED ISSUE #126: Add lzf_filter
Some .h5 files require `liblzf_filter.so`. I git-cloned h5py, compile `liblzf_filter` and installed it in `/usr/local/hdf5/lib/plugin/`
Whether I use `hdf5-sys/static` or not, it complains that 'lzf' is not registered.

mulimoen:
This is a continued effort on the `dcpl`-branch of @aldanor. This adds a lot more functionality and makes the API much nicer to use for selections.

See mulimoen#2 for some earlier comments and discussions regarding this branch.

Fixes #45
Fixes #88
Fixes #126


TheButlah:
This PR looks super awesome, but is it abandoned now?

aldanor:
> This PR looks super useful, but is it abandoned now?

In brief: nope.

CHANGELOG.md
```diff
@@ -2,6 +2,23 @@
...
+
+### Changed
+- `fill_value` now returns `DynValue`.
+- `hdf5-types` conditionally uses the `hdf5` allocator when necessary.
+- Compression filters has been renamed.
+- The automatic chunking uses a fill-from-back approach instead of the
+  approach used by `h5py`.
+- `DatasetBuilder` now comes in different flavours to reflect additional
+  information used during the building of a dataset.
+
```
aldanor, line 21: # Added

- Complete rewrite of `DatasetBuilder`; dataset creation API is now different and not
  backwards-compatible (however, it integrates all of the new features and is more
  flexible and powerful). It is now possible to create and write the datasets in one step.
  Refer to the API docs for full reference.
- New `Extents` type matching HDF5 extents types: null (no elements), scalar, simple (fixed
  dimensionality); it is used to query and specify shapes of datasets. Extents objects
  are convertible from numbers and also tuples, slices and vectors of indices - all of
  which can be used whenever extents are required (e.g., when creating a new dataset).
- New `Selection` type and the API around it closely matching HDF5 selection API - this
  includes 'all' selection, point-wise selection and hyperslab selection (only 'regular'
  hyperslabs are supported - that is, hyperslabs that can be represented as a single
  multi-dimensional box some of whose dimensions may be infinite). Selection objects
  are convertible from numbers and ranges and tuples and arrays of numbers and ranges;
  one can also use `s!` macro from `ndarray` crate. Selections can be provided when
  reading and writing slices.
- LZF / Blosc filters have been added, they have to be enabled by `"lzf"` / `"blosc"`
  cargo features and depend on `lzf-sys` / `blosc-sys` crates respectively. Blosc filter
  is a meta-filter providing multi-threaded access to the best-in-class compression codecs
  like Zstd and LZ4 and is recommended to use as a default when compression is critical.
- New `Filter` type to unify all of the filters API; if LZF / Blosc filters are enabled,
  this enum also contains the corresponding variants. It is now also possible to provide
  user-defined filters with custom filter IDs and parameters.
- Dataset creation property list (DCPL) API is now supported; this provides access to all
  of the properties that can be specified at dataset creation time (e.g., layout, chunking,
  fill values, external file linking, virtual maps, object time tracking, attribute
  creation order, and a few other low-level settings).
- As part of DCPL change, virtual dataset maps (VDS API in HDF5 1.10+) are now supported.
- Link creation property list (LCPL) API is now also wrapped.
- File creation property list (FCPL) API has been extended to include a few previously
  missing properties (object time tracking, attribute creation order and few other
  low-level settings).
- Added `const-generics` feature to `hdf5-types` crate - adds const generics support for
  Rust 1.51 (will be eventually added to default features list).
- Added `force-h5-allocator` feature to `hdf5-types` crate - forces HDF5 allocator for
  varlen types and dynamic values. This may be necessary on platforms where different
  allocators may be used in different libraries (e.g. dynamic libraries on Windows),
  or if `libhdf5` is compiled with the memchecker option enabled. This option is
  force-enabled by default if using a dll version of the library on Windows.
- New `DynValue` type which represents a dynamic self-describing HDF5 object that
  also knows how to deallocate itself; it supports all of the HDF5 types including
  compound types, strings and arrays.
- Added methods to `Dataset`: `layout`, `dapl`, `access_plist`, `dcpl`, `create_plist`.
  
  ** Changed
  
- `Dataspace` type has been reworked and can be now constructed from an extents object 
  and sliced with a selection object.
- `Dataset::fill_value` now returns an object of the newly added `DynValue` type; this
  object is self-describing and knows how to free itself.
- Automatic chunking now uses a fill-from-back approach instead of the previously
  used method which is used in `h5py`.
- Removed `Filters` type (there's now `Filter` that represents a single filter).
- `write_slice`, `read_slice`, `read_slice_1d`, `read_slice_2d` now take any object
  convertible to `Selection` (instead of `SliceInfo`).
- `Dataset::chunks` has been renamed to `Dataset::chunk`

CHANGELOG.md
```diff
@@ -2,6 +2,23 @@
...
+- Dataset creation property lists (DCPL)
+
+### Changed
+- `fill_value` now returns `DynValue`.
+- `hdf5-types` conditionally uses the `hdf5` allocator when necessary.
+- Compression filters has been renamed.
+- The automatic chunking uses a fill-from-back approach instead of the
+  approach used by `h5py`.
+- `DatasetBuilder` now comes in different flavours to reflect additional
+  information used during the building of a dataset.
```
aldanor, line 20:
```suggestion
### Added

- Complete rewrite of `DatasetBuilder`; dataset creation API is now different and not
  backwards-compatible (however, it integrates all of the new features and is more
  flexible and powerful). It is now possible to create and write the datasets in one step.
  Refer to the API docs for full reference.
- New `Extents` type matching HDF5 extents types: null (no elements), scalar, simple (fixed
  dimensionality); it is used to query and specify shapes of datasets. Extents objects
  are convertible from numbers and also tuples, slices and vectors of indices - all of
  which can be used whenever extents are required (e.g., when creating a new dataset).
- New `Selection` type and the API around it closely matching HDF5 selection API - this
  includes 'all' selection, point-wise selection and hyperslab selection (only 'regular'
  hyperslabs are supported - that is, hyperslabs that can be represented as a single
  multi-dimensional box some of whose dimensions may be infinite). Selection objects
  are convertible from numbers and ranges and tuples and arrays of numbers and ranges;
  one can also use `s!` macro from `ndarray` crate. Selections can be provided when
  reading and writing slices.
- LZF / Blosc filters have been added, they have to be enabled by `"lzf"` / `"blosc"`
  cargo features and depend on `lzf-sys` / `blosc-sys` crates respectively. Blosc filter
  is a meta-filter providing multi-threaded access to the best-in-class compression codecs
  like Zstd and LZ4 and is recommended to use as a default when compression is critical.
- New `Filter` type to unify all of the filters API; if LZF / Blosc filters are enabled,
  this enum also contains the corresponding variants. It is now also possible to provide
  user-defined filters with custom filter IDs and parameters.
- Dataset creation property list (DCPL) API is now supported; this provides access to all
  of the properties that can be specified at dataset creation time (e.g., layout, chunking,
  fill values, external file linking, virtual maps, object time tracking, attribute
  creation order, and a few other low-level settings).
- As part of DCPL change, virtual dataset maps (VDS API in HDF5 1.10+) are now supported.
- Link creation property list (LCPL) API is now also wrapped.
- File creation property list (FCPL) API has been extended to include a few previously
  missing properties (object time tracking, attribute creation order and few other
  low-level settings).
- Added `const-generics` feature to `hdf5-types` crate - adds const generics support for
  Rust 1.51 (will be eventually added to default features list).
- Added `force-h5-allocator` feature to `hdf5-types` crate - forces HDF5 allocator for
  varlen types and dynamic values. This may be necessary on platforms where different
  allocators may be used in different libraries (e.g. dynamic libraries on Windows),
  or if `libhdf5` is compiled with the memchecker option enabled. This option is
  force-enabled by default if using a dll version of the library on Windows.
- New `DynValue` type which represents a dynamic self-describing HDF5 object that
  also knows how to deallocate itself; it supports all of the HDF5 types including
  compound types, strings and arrays.
- Added methods to `Dataset`: `layout`, `dapl`, `access_plist`, `dcpl`, `create_plist`.
  
 ### Changed
  
- `Dataspace` type has been reworked and can be now constructed from an extents object 
  and sliced with a selection object.
- `Dataset::fill_value` now returns an object of the newly added `DynValue` type; this
  object is self-describing and knows how to free itself.
- Automatic chunking now uses a fill-from-back approach instead of the previously
  used method which is used in `h5py`.
- Removed `Filters` type (there's now `Filter` that represents a single filter).
- `write_slice`, `read_slice`, `read_slice_1d`, `read_slice_2d` now take any object
  convertible to `Selection` (instead of `SliceInfo`).
- `Dataset::chunks` has been renamed to `Dataset::chunk`
```

aldanor: @mulimoen I've tried including most of the things I could find that were changed - maybe there's something else I've missed?

(after we merge this, I think what we should do is start working on the docs, right now there's so many types and APIs and only a small fraction is documented so in many cases it would be completely unclear how to use them... ideally docs.rs documentation should have a good quickstart guide as well)

.github/workflows/ci.yml
```diff
@@ -164,7 +164,7 @@ jobs:
           cargo test -vv --features="$FEATURES"
       - name: Test const generics
         if: matrix.rust == 'nightly'
-        run: cargo test -p hdf5-types --features hdf5-types/const_generics
+        run: cargo test -p hdf5-types --features hdf5-types/const-generics
```
aldanor, line 167: Now that 1.51 is out, I wonder if we should just set MSRV to 1.51, or wait a bit? Hmm..

Crates like arrayvec now require 1.51 (with support for const generics). ndarray is moving slowly in that direction, it's likely that the next major release will require 1.51 (to happen in the foreseeable future).

hdf5-types/src/lib.rs
```diff
@@ -4,20 +4,57 @@
 //! Types that can be stored and retrieved from a `HDF5` dataset
 //!
 //! Crate features:
-//! * `const_generics`: Uses const generics to enable arrays [T; N] for all N.
+//! * `const-generics`: Uses const generics to enable arrays [T; N] for all N.
 //!                     Compiling without this limits arrays to certain prespecified
 //!                     sizes
+//! * `force-h5-allocator`: Force the `hdf5` allocator for varlen types and dynamic values.
```
aldanor, line 10: Can we rename this feature to `h5-allocator` or even simply `h5-alloc`? (the changelog to be updated as well in this case)

Also, should there be a matching feature in the root crate, or not?

hdf5-types/Cargo.toml
```diff
@@ -8,13 +8,18 @@ description = "Native Rust equivalents of HDF5 types."
...
 
 [features]
-const_generics = []
+const-generics = []
+force-h5-allocator = []
 
 [dependencies]
 ascii = "1.0"
 libc = "0.2"
+hdf5-sys = { version = "0.7.1", path = "../hdf5-sys" }
```
aldanor, line 20: Add a `# !V` here? Try grepping for `0.7.1`, all lines with it should have this tag. There's a shell script that helps bumping all the versions at once.

hdf5-types/Cargo.toml
```diff
@@ -8,13 +8,18 @@ description = "Native Rust equivalents of HDF5 types."
...
 edition = "2018"
+build = "build.rs"
 
 [features]
-const_generics = []
+const-generics = []
+force-h5-allocator = []
 
 [dependencies]
 ascii = "1.0"
```
aldanor, line 18: As for various 3rd party scripts - try running `cargo outdated -R -w` to see if some versions can/need to be bumped? (since we're already changing a lot)

CHANGELOG.md
mulimoen: I think we've added all the major items now.  Yes, we need a much more comprehensive documentation. We have some documentation, but it split between the documentation on `docs.rs` and in the examples. A quick-start would be a great idea, with some links to external resources on how to use `hdf5` the most efficiently.

hdf5-types/src/lib.rs
mulimoen: I'll rename this to `h5-alloc`. I think leaving this to `hdf5-sys` is alright. We havea presedence in the `static` feature not being exposed from `root`.

aldanor: 👍

aldanor:
Great, thanks. I think I've finished skimming over the code in general - let me re-read the actual dataset / dataset-builder part today and I'll send some comments on that if any. I think after that we're good to go.

(I wouldn't make a release before probably merging in #139 too since it's quite a big change - will get back re: that as well)

.github/workflows/ci.yml
mulimoen: `ndarray` version 0.15 has 1.49 as their official MSRV. 1.51 is only a couple of months after this. Having the `const-generics` flag only makes sense for two compiler versions...

src/hl/selection.rs
```diff
@@ -0,0 +1,1476 @@
...
+
+#[allow(clippy::fallible_impl_from)]
+impl<T: Into<ndarray::SliceInfoElem>> From<T> for SliceOrIndex {
+    fn from(slice: T) -> Self {
+        match slice.into() {
+            ndarray::SliceInfoElem::Index(index) => Self::Index(index),
+            ndarray::SliceInfoElem::Slice { start, end, step } => {
+                Self::Slice { start, step, end, block: false }
+            }
+            ndarray::SliceInfoElem::NewAxis => panic!("ndarray NewAxis can not be mapped to hdf5"),
```
mulimoen, line 278: This is a bit troublesome for conversion from the new `SliceInfo` struct. `ndarray` is OK with panicing on indexing failures, but we could try to use the `TryFrom` trait everywhere instead. Maybe something to fix at a later time?

.github/workflows/ci.yml
mulimoen: I've removed the flag in a commit, but we can revert this if we want to keep it.

CHANGELOG.md
```diff
@@ -59,6 +57,7 @@
 - `write_slice`, `read_slice`, `read_slice_1d`, `read_slice_2d` now take any object
   convertible to `Selection` (instead of `SliceInfo`).
 - `Dataset::chunks` has been renamed to `Dataset::chunk`
+- `hdf5-types` uses `const-generics` for arrays, allowing arbitrary sized arrays.
```
aldanor, line 60: Probably worth mentioning in the changelog that MSRV is now 1.51?

src/hl/selection.rs
aldanor: Yea... also:

- `impl<T, Din, Dout> From<ndarray::SliceInfo<T, Din, Dout>> for Hyperslab` 
  relies on `From<ndarray::SliceIndex>` so it would have to become a `TryFrom` as well
- Whatever uses that will have to become a `TryFrom`
- `impl From<Array1<Ix>> for Selection` also contains an `.unwrap()`

It's not nice having a core part of API potentially panicking whereas we return `Result<>` almost everywhere so we can handle it. Probably worth reviewing this entire module to see if we can avoid all panics (e.g. by using `TryFrom`).

licenses/lzf.txt
```diff
@@ -0,0 +1,33 @@
+Copyright Notice and Statement for LZF filter
```
aldanor, line 1: I wonder now if we have to include licenses for lzf/blosc/h5py: because we no longer include C sources for lzf/blosc (which, I think, was the case early on) - as we rely on blosc-sys and lzf-sys.

CHANGELOG.md
aldanor: Given that `const-generics` is no longer a crate feature, perhaps rewrite this as?

```markdown
- Const generics support (Rust 1.51): `hdf5-types` now uses const generics for array types, 
  allowing fixed-size arrays of arbitrary sizes.
```

hdf5-types/src/array.rs
```diff
@@ -14,76 +14,22 @@ pub unsafe trait Array: 'static {
     fn capacity() -> usize;
 }
```
aldanor, line 16: I think this trait `Array` is now not needed at all. The idea was initially borrowed from `arrayvec` which now also switched to const generics - see https://github.com/bluss/arrayvec/pull/172 (and the removed `array.rs`)?

Cargo.toml
```diff
@@ -24,21 +26,25 @@ bitflags = "1.2"
 lazy_static = "1.4"
 libc = "0.2"
 parking_lot = "0.11"
-ndarray = ">=0.13,<0.15"
+ndarray = "0.15.1"
```
aldanor, line 29: Is 0.15.1 incompatible with 0.15, why not `=0.15`?

hdf5-types/Cargo.toml
```diff
@@ -8,13 +8,17 @@ description = "Native Rust equivalents of HDF5 types."
...
 
 [features]
-const_generics = []
+h5-alloc = []
 
 [dependencies]
 ascii = "1.0"
 libc = "0.2"
+hdf5-sys = { version = "0.7.1", path = "../hdf5-sys" }  # !V
+cfg-if = "1.0.0"
```
aldanor, line 20: Simply `=1.0`

src/hl/dataset.rs
```diff
@@ -140,728 +168,1112 @@ impl Dataset {
...
-                    let mut value = mem::MaybeUninit::<T>::uninit();
-                    h5try!(
-                        H5Pget_fill_value(dcpl_id, datatype.id(), value.as_mut_ptr() as *mut _,)
-                    );
-                    Ok(Some(value.assume_init()))
-                }
+            let dtype_src = Datatype::from_type::<T>()?;
+            let dtype_dst = Datatype::from_descriptor(&self.type_desc)?;
+            // TODO: soft conversion? hard? user-specifiable?
+            dtype_src.ensure_convertible(&dtype_dst, Conversion::Soft)?;
```
aldanor, line 335: Conversion: this would be quite easy to do, no? By just adding the same API to the builder as in `Container`:

```rust
    /// Set maximum allowed conversion level.
    pub fn conversion(mut self, conv: Conversion) -> Self {
        self.conv = conv;
        self
    }

    /// Disallow all conversions.
    pub fn no_convert(mut self) -> Self {
        self.conv = Conversion::NoOp;
        self
    }
```

src/hl/dataset.rs
```diff
@@ -140,728 +168,1112 @@ impl Dataset {
...
+    pub fn create_intermediate_group(&mut self, create: bool) {
+        self.with_lcpl(|pl| pl.create_intermediate_group(create));
     }
+
+    pub fn char_encoding(&mut self, encoding: CharEncoding) {
+        self.with_lcpl(|pl| pl.char_encoding(encoding));
+    }
+}
+
+macro_rules! impl_builder_stuff {
```
aldanor, line 819: Wonder - are the generated docs ok with these macros?

aldanor:
@mulimoen I've went through the code once again and (the lack of docs and docstrings aside, and also the complete lack of tests aside) it's pretty good to go. I left a few minor comments here and there.

(I really think we should write some docs and tests after this is released and is considered stable api-wise.)

CHANGELOG.md
mulimoen: #141

src/hl/dataset.rs
mulimoen: The docs are replicated on all relevant types, this is duplication, but it all shows as a separate impl block. I've added a small comment on the `impl` blocks to describe how they are all common between dataset-builders

src/hl/dataset.rs
mulimoen: Done

mulimoen:
I've opened some new GH issues for some of the remaining tasks. This PR is enourmous as it is, and the other issues can be merged and reviewed separately.

licenses/lzf.txt
mulimoen: I think the rework of automatic chunking also means we now no longer have any direct code from `h5py`. We also comply with the `hdf5` license by including this in the `hdf5-src` archive. We should be safe to remove `licenses` (but I am not a lawyer).

aldanor: Yes, I think they are safe to remove now.

aldanor:
Good point re: separate tickets (array trait, docs, avoiding panics), they can be handled separately. Maybe also a ticket for adding full tests for `dataset` and all related builders etc? A lot of the code in this PR is fully covered by tests (independent parts like extents, selection, dyn value, etc). What needs to be thoroughly tested is all of the new builders and the new Dataset - also lzf/blosc compression and all the new features.

aldanor:
At this point I'm somewhat scared to press the green button because this PR is so big my browser sometimes hangs when viewing diffs :) Great work @mulimoen in resurrecting it and finishing all the unfinished parts!

(Before the release, as I've said, I think we should consider merging at least the ownership/IDs PR)

aldanor (approved):

aldanor:
(Added a note that this fixes #47 as well)

mulimoen:
> (Added a note that this fixes #47 as well)

Only halfway, one would need to dispatch based on some fixed sizes.

aldanor:
> > (Added a note that this fixes #47 as well)
> 
> Only halfway, one would need to dispatch based on some fixed sizes.

Hm? The question was to create a dataset with runtime-known string size. This is now doable via `empty_as()` and `with_data_as()`?

mulimoen:
I think what @pmarks is trying is creating the type from an `n` that is only known at runtime (max of all strings), but one still needs to define this type at compile time to use at runtime. The workaround in https://github.com/aldanor/hdf5-rust/issues/47#issuecomment-641915316 is using a giant `match` to work around this by letting each arm have the compile time fixed value. I'll remove the issue so it's not closed on merge.

mulimoen:
Thank you very much for reviewing and finishing this @aldanor! Hopefully future PRs won't be as massive as this one!

aldanor:
Yea - well we solved it halfway by allowing runtime descriptors, that's already something that wasn't previously possible, so that's something.

Also, that 'giant match' can now be implemented via const generics.

Finally, with strings... I have a ticket with a hacky solution for Rust types like strings - maybe we'll get to that at some point as well.

aldanor:
> Thank you very much for reviewing and finishing this @aldanor! Hopefully future PRs won't be as massive as this one!

Yea, let's try to keep it simple from now on :) I may open a few PRs in the nearby future with some docs updates.

PR #146, opened 2021-04-08
Remove num-dependencies

### LINKED ISSUE #135: Consider removing num-traits and num-integer deps
These are not heavyweight crates but still, would be nice to trim down deps list.

`num_integer` and `num_traits` are only used in `util` and `error` modules to handle errors of various types, whereas in fact there may only be one (`herr_t`) or two integer types used for error return codes. We could simply add our own `ErrorCode` trait for those types (or use `herr_t` everywhere, and convert to it if it's a different type).

`num_integer::div_floor` is used in `infer_chunk_size()`, but that's just out of convenience, can be easily replaced with `/` and `%`.

mulimoen:
Fixes #135 now that the automatic chunking no longer uses num-integer.

Implementing the trait for the whitelisted types helped find a bug in turning a compound datatype into a descriptor.


src/hl/datatype.rs
```diff
@@ -261,7 +261,7 @@ impl Datatype {
                     let mut fields: Vec<CompoundField> = Vec::new();
                     for idx in 0..h5try!(H5Tget_nmembers(id)) as _ {
                         let name = H5Tget_member_name(id, idx);
-                        let offset = h5try!(H5Tget_member_offset(id, idx));
+                        let offset = H5Tget_member_offset(id, idx);
```
aldanor, line 264: Hmm...

> Returns the byte offset of the field if successful; otherwise returns 0 (zero). Note that zero is a valid offset and that this function will fail only if a call to H5Tget_member_class() fails with the same arguments.

Indeed, looks like it can't fail. If there's something wrong, then it's the line below (assigning `ty`) that will fail.

src/util.rs
```diff
@@ -59,9 +56,9 @@ pub fn h5_free_memory(mem: *mut c_void) {
 pub fn get_h5_str<T, F>(func: F) -> Result<String>
 where
     F: Fn(*mut c_char, size_t) -> T,
-    T: Integer + NumCast,
+    T: std::convert::TryInto<isize>,
```
aldanor, line 59: Minor nit: maybe `use std::convert::TryInto` above?

src/error.rs
```diff
@@ -287,25 +285,51 @@ impl From<ShapeError> for Error {
...
-{
-    if is_err_code(value) {
-        Error::query().map_or_else(|| Ok(value), Err)
-    } else {
-        Ok(value)
+impl H5ErrorCode for hsize_t {
+    fn is_err_code(value: Self) -> bool {
+        value == 0
+    }
+}
```
aldanor, line 312: Minor: newline before each `impl`?

src/error.rs
```diff
@@ -287,25 +285,51 @@ impl From<ShapeError> for Error {
...
+    H5ErrorCode::h5check(value)
+}
+
+#[allow(unused)]
+pub fn is_err_code<T: H5ErrorCode>(value: T) -> bool {
+    H5ErrorCode::is_err_code(value)
+}
+
+pub trait H5ErrorCode: Copy {
+    fn is_err_code(value: Self) -> bool;
```
aldanor, line 298: Minor: newline after the first fn?

src/error.rs
```diff
@@ -287,25 +285,51 @@ impl From<ShapeError> for Error {
...
+    }
+}
+#[cfg(hdf5_1_10_0)]
+impl H5ErrorCode for hid_t {
+    fn is_err_code(value: Self) -> bool {
+        value < 0
+    }
+}
+#[cfg(not(hdf5_1_10_0))]
+impl H5ErrorCode for hdf5_sys::h5::hssize_t {
```
aldanor, line 325: `use hdf5_sys::h5::hssize_t` above?

CHANGELOG.md
```diff
@@ -46,7 +46,8 @@
 - Added methods to `Dataset`: `layout`, `dapl`, `access_plist`, `dcpl`, `create_plist`.
   
  ### Changed
-  
+
+- Required Rust compiler version is now `1.51`.
```
aldanor, line 50: Maybe add a 

```markdown
- Removed `num-integer` and `num-traits` dependencies.
```

(but leave the 1.51 line the first so it stands out)

src/hl/datatype.rs
mulimoen: Exactly the same for `H5Tget_size`, but I could not make a test with a zero-sized type. Tried with a `[0; u8]`, but empty fixed arrays are not legal.

src/error.rs
mulimoen: Would require a `cfg` on that line too

aldanor: That's ok, I think, to make it inline with other modules where there's tons of cfg imports

aldanor (approved):

aldanor:
Looks great, thanks!

PR #147, opened 2021-04-10
Use TryFrom over From

### LINKED ISSUE #142: Remove potential panics in From impl
The current signature of the `Container` object requires extents which implements `Into<Extents>`. This should be modified to accept `TryInto<Extents>`, and implementations of `From<Foo> for Extents` should be changed into `TryFrom`, as `From`/`Into` should never panic.

See https://github.com/aldanor/hdf5-rust/pull/140#discussion_r608241744

mulimoen:
This changes the `From` implementations of `SliceOrIndex`/`Hyperslab`/`Selection` to `TryFrom`.

As we now manually implement most methods instead of using `ndarray`s `Into<SliceInfoElem>`, would it make sense to change our `SliceOrIndex` to instead match `hdf5` semantics, and recommend using tuples for indexing? `SliceOrIndex` would match better if they had members of `Ix` instead of `isize`, we are converting everything to `RawSelection` which would fail anyways.

If we introduce a macro for setting the step/block we would have feature parity with `ndarray::s`. Syntax for this is up for discussion.

Fixes #142


mulimoen:
This change is orthogonal to all other PRs. There is need for changes in the CHANGELOG, this overwrites the unreleased `Extents` and friends

mulimoen:
@aldanor Rebased and green

PR #154, opened 2021-05-20
Update hdf5-src to use 1.10.7

### LINKED ISSUE #153: Cannot compile hdf5-src on macos
It fails with
```shell
H5Fsuper.c:1397:16: error: implicit declaration of function
'H5O__fsinfo_set_version' is invalid in C99
[-Werror,-Wimplicit-function-declaration]
             if(H5O__fsinfo_set_version(f, &fsinfo) < 0)
```
The solution is to add 
```
-D CMAKE_C_FLAGS="-Wno-error=implicit-function-declaration"
```
to cmake.

mulimoen:
This version is mostly a maintaince release, but introduces some new functions for allocation information which might be of use.

Fixes #153

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,9 +126,6 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
         with: {submodules: true}
-      - name: Change to older toolchain
-        if: matrix.os == 'macos'
-        run: sudo xcode-select -s "/Applications/Xcode_11.7.app"
       - name: Install Rust (${{matrix.rust}})
         uses: actions-rs/toolchain@v1
         with: {toolchain: '${{matrix.rust}}', profile: minimal, override: true}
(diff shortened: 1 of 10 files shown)


aldanor (approved):

aldanor:
Changelog? (for both adding the apis and for changing the default version for hdf5-src)

aldanor:
> Merge gone wrong, fixed

👍

mulimoen:
This is a bit tricky without a windows computer... I think the `libz` that is built has a different name on `msvc`, but I am not sure how to find this. I've opened an issue (https://github.com/rust-lang/libz-sys/issues/80) to get this name through env inspection instead.

mulimoen:
Turns out a `hdf5-c` utility was being built, and tried to link to `libz`. Why this did not fail on other platforms is a mystery. Perhaps all other platforms includes some `libz` by default?

PR #157, opened 2021-06-04
Add more iterators on Group

mulimoen:
This adds `groups`, `datasets`, and `datatypes` to `Group`. This also adds the general `iter_visit`, which can be adapted by users to fully use the capabilities of `H5Literate` (sans iteration resumption).


mulimoen:
@aldanor This should be good to go now

aldanor:
@mulimoen Looks good overall. I started writing review but figured it would be easier to fix myself - I'll open a PR to your PR on your fork.

mulimoen:
`hdf5` has ownership of this ID, and we need to ensure we don't invalidate the id before we return it, as there is an unconditional close on the ID after the callback

mulimoen:
@aldanor Looks good to me now. I forgot how big this release is, I thought we already published attributes, but apparently not

aldanor:
Hm, with times not being filled out, that's not good - but I'll open a separate PR to quickly fix that. This one - let's merge then.

Issue #155, opened 2021-05-20
Silence errors by default, allow unsilencing them?


aldanor:
@mulimoen I had this idea for a while now, and most libraries actually do the same:

Errors should be silenced by default, and can be unsilenced purposefuly, e.g. when debugging something. We capture the entire error stack anyway on every error, so there's no need to dump stuff to console at the same time.

(also look at the readme example! would be nice to remove it from there)

mulimoen:
There is tons of functionality regarding errors, we can definitely be a bit smarter about this, e.g.: 
* Extend the error type to grab the error stack on error, and let the user deal with the backtrace (this we partially do already)
* Register a class to handle errors (forward to `log`?)

aldanor:
The first one we kind of already do (although it's very expensive because of the backtrace extraction; so if there's any logic relying on catching errors it will be very slow).

Adding an error handler sounds pretty cool, but you could do it yourself on the outside, no? If you have a `Result<>`, you can choose if you want to forward the error to log or not, yourself?

Whichever the case, I think that *by default* the errors should be silenced, unlike the current default.

mulimoen:
I fully agree errors should be silenced by default, should be easy to hook up the `/dev/zero` handler in `LIBRARY_INIT`. I think a PR focusing on this should addiionally simplify `ErrorStack` a bit, instead of expanding it would be cheaper to pop off the error object (the hid) and expand at a later time.

Rikorose:
@mulimoen in multi-threading scenarios, errors are still shown. This can be also seen e.g. during cargo test as cargo runs the tests in multiple threads.

Is there a way to silence the errors also when using multiple threads to access an hdf5 dataset?

mulimoen:
@Rikorose The current stack is a thread local variable, and the error printer is not set for all error stacks. Could you try calling `sync(||{})` before spawning threads and check if that silences the errors? Could be that we have to check TLS and call `silence_errors` for each thread and not just on the first we encounter

Rikorose:
Does not seem to work. I am using the global rayon thread pool for multi threadding:
```rust
hdf5::sync::sync(|| {});
let mut poolbuilder = rayon::ThreadPoolBuilder::new();
poolbuilder = poolbuilder.start_handler(|_| hdf5::sync::sync(|| {}));
poolbuilder.build_global()?;
```
I am still seeing errors like:
```
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 139802791892736:
[...]

PR #158, opened 2021-06-10
Export SilenceErrors type

Rikorose:
This allows storing it in places where we need a type annotation.


mulimoen:
See also #155 where we consider obsoleting the `SilenceError` struct

Rikorose:
Ah yes, this would also be an option.

PR #161, opened 2021-07-04
Leave the error stack unexpanded

### LINKED ISSUE #155: Silence errors by default, allow unsilencing them?

mulimoen:
This introduces a wrapper around the errorstack, which should make the error stack a bit easier to reason about. Errors are now silenced by default

Fixes #155
Closes #158

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -18,6 +18,80 @@ use hdf5_sys::h5e::{
 
 use crate::internal_prelude::*;
 
+pub mod h5 {
+    use super::*;
+    #[repr(transparent)]
+    #[derive(Clone)]
+    pub struct ErrorStack(Handle);
+
+    impl ObjectClass for ErrorStack {
+        const NAME: &'static str = "errorstack";
+        const VALID_TYPES: &'static [H5I_type_t] = &[H5I_ERROR_STACK];
+
+        fn from_handle(handle: Handle) -> Self {
+            Self(handle)
+        }
+
+        fn handle(&self) -> &Handle {
+            &self.0
+        }
+
+        // TODO: short_repr()
+    }
+
+    impl ErrorStack {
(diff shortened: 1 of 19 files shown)


src/error.rs
```diff
@@ -51,141 +117,36 @@ impl ErrorFrame {
...
-
 #[derive(Clone, Debug)]
-pub struct ErrorStack {
+pub struct ExpandedErrorStack {
     frames: Vec<ErrorFrame>,
     description: Option<String>,
 }
 
-impl Index<usize> for ErrorStack {
+impl Index<usize> for ExpandedErrorStack {
```
aldanor, line 126: Thinking about, could probably `Deref<[ErrorFrame]>`?

src/globals.rs
```diff
@@ -23,6 +23,7 @@ lazy_static! {
             // still be live on other threads on program exit
             ::hdf5_sys::h5::H5dont_atexit();
             ::hdf5_sys::h5::H5open();
+            ::hdf5_sys::h5e::H5Eset_auto2(::hdf5_sys::h5e::H5E_DEFAULT, None, core::ptr::null_mut());
```
aldanor, line 26: The question is now though, do we want to provide a way to insolence errors? Perhaps a fancy mutex counter thing is not needed anymore, more like a simple global 'silence_errors(true/false)`?

aldanor:
Looks awesome, thanks!

aldanor (approved):

mulimoen:
Thanks for the review @aldanor!

PR #163, opened 2021-07-11
Don't unload hdf5 in build script

### LINKED ISSUE #162: CI failures on macos/brew/1.12
It seems the current brew 1.12 using mpi for macos is acting flaky, and will segfault in a build script on building `hdf5-sys`.

An update of the homebrew repo means we now pull in `hwloc` version 2.5.0 (always failing), instead of version 2.4.0 (always working), which seems to be the only difference in installed packages. This could be relevant for debugging this issue.

mulimoen:
The build failures described in #162 were most likely due to [TLS destructors behaving poorly with dynamic loading](https://github.com/nagisa/rust_libloading/issues/5). We work around this by leaving the library unloaded, which should not have any bad effects when used in the build script. `h5_dontatexit` did not have any effect on this bug.

Fixes #162


mulimoen:
@aldanor If you merge this you won't be spammed by the failing CI runs anymore

aldanor (approved):
Wow, thanks for digging to the bottom of this! 👍

PR #166, opened 2021-08-16
Fix compiler warnings and clippy lints

mulimoen:
@aldanor This fixes the latest CI failures


aldanor:
Thanks @mulimoen, as always :)

// I'm back next week after a few month long vacation trip and I'm keen on jumping back on train so we can release the big version and move on forward, e.g. with docs etc.

aldanor (approved):

PR #175, opened 2021-10-16
Fix deadlock on init

### LINKED ISSUE #174: Race condition in initialisation
The following two tests will often cause a hang on `cargo test` when started at the same time
```rust
    #[test]
    fn lock_part1() {
        h5call!(*H5P_ROOT);
    }
    #[test]
    fn lock_part2() {
        *H5P_ROOT;
    }
```

mulimoen:
This should always initialise the library correctly without dead-locking when nesting `LIBRARY_INIT` and `LOCK`

Fixes #174


aldanor:
That's great! Yea, that should fix it. Maybe update the changelog as well (in the 'fixed' section)? Or is it sometime we have introduced past 0.7.1?

(Few cosmetic comments aside, looks good to go if it's all green)

aldanor:
👍 Great. I'll rebase my loc branch on that then and push it shortly

Issue #141, opened 2021-04-07
Add documentation


mulimoen:
A lot of the items and functions in this crate are undocumented, and it is hard to understand how to use this crate (and hdf5) in the correct way for maximum performance.

We should minimally add documentation for most types and functions, to make the documentation on `docs.rs` as clear as possible. Doctests can be added to individual functions to illustrate basic principals.

Having a guide on how to use the crate to fit in with the `hdf5` ecosystem would be nice. We could link the current examples, and add additional one for how to use chunking, performance pitfalls (compression on varlen arrays, SOA vs AOS), and related functionality/tips & tricks in `h5py` and other relevant locations.

mulimoen:
It would be great to include examples into our docs as per https://twitter.com/wcrichton/status/1456112165744615426

robgjansen:
FWIW, I find myself wanting to use this crate but the lack of documentation keeps driving me back to python's h5py library. The idea to create a book along the lines of h5py is a great one and would make it much easier to get up to speed quickly.

PR #181, opened 2021-10-24
Experimental docs.rs changes

aldanor:
1. Enable `doc_cfg` for nightly builds (on docs.rs and locally if enabled)
2. (Experimental) Replace `hdf5_1_2_3` config flags with `feature = "1.2.3"`. This will produce annotations like "Enabled on **crate feature** 1.10.0 only" in the docs instead of "Enabled on `hdf5_1_10_0` only". If I understand correctly, this will also allow the users to refer to these as `hdf5/1.10.0` externally (?). Features can be named as "hdf5-1.10.0" but that would look a bit ugly in the docs (with the numbers only, it looks pretty neat, see below for examples).
3. (TODO) we should definitely build the docs on the latest HDF5 version available (1.12.1?).
4. (TODO) ideally, we want mpio/blosc/lzf on docs.rs as well (the latter two are already there).
5. Should `h5_have_parallel` and such be renamed to `feature = "have-parallel"`?

Note: support for period for feature names has been only recently added to cargo (circa 1.50).

Related: #141.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,10 @@ default = []
 mpio = ["mpi-sys", "hdf5-sys/mpio"]
 lzf = ["lzf-sys", "errno"]
 blosc = ["blosc-sys"]
+# The features with version numbers such as 1.10.3, 1.12.0 are metafeatures
+# and is only available when the HDF5 library is at least this version.
+# Features have_direct and have_parallel are also metafeatures and dependent
+# on the HDF5 library which is linked against.
 
 [workspace]
 members = [".", "hdf5-types", "hdf5-derive", "hdf5-sys", "hdf5-src"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,3 +47,4 @@ tempfile = "3.2"
 
 [package.metadata.docs.rs]
 features = ["hdf5-sys/static", "hdf5-sys/zlib", "blosc", "lzf"]
+rustdoc-args = ["--cfg", "docsrs"]
--- a/build.rs
+++ b/build.rs
@@ -1,19 +1,21 @@
 use std::env;
 
 fn main() {
+    let print_feature = |key: &str| println!("cargo:rustc-cfg=feature=\"{}\"", key);
+    let print_cfg = |key: &str| println!("cargo:rustc-cfg={}", key);
     for (key, _) in env::vars() {
-        let key = match key.as_str() {
-            "DEP_HDF5_HAVE_DIRECT" => "have_direct".into(),
-            "DEP_HDF5_HAVE_STDBOOL" => "have_stdbool".into(),
-            "DEP_HDF5_HAVE_PARALLEL" => "have_parallel".into(),
-            "DEP_HDF5_HAVE_THREADSAFE" => "have_threadsafe".into(),
-            "DEP_HDF5_MSVC_DLL_INDIRECTION" => "dll_indirection".into(),
+        match key.as_str() {
+            // public features
+            "DEP_HDF5_HAVE_DIRECT" => print_feature("have-direct"),
+            "DEP_HDF5_HAVE_PARALLEL" => print_feature("have-parallel"),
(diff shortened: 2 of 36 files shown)


Cargo.toml
```diff
@@ -16,6 +16,10 @@ default = []
 mpio = ["mpi-sys", "hdf5-sys/mpio"]
 lzf = ["lzf-sys", "errno"]
 blosc = ["blosc-sys"]
+# The features with version numbers such as 1.10.3, 1.12.0 are metafeatures
+# and is only available when the HDF5 library is at least this version.
+# Features have_direct and have_parallel are also metafeatures and dependent
```
aldanor, line 21: @mulimoen `have_direct` and `have_parallel` and not features as of now. Should we make them features as well? I.e., `feature = "have-direct"` and `feature = "have-parallel"`, so that they show up the same way as versions?

build.rs
```diff
@@ -3,14 +3,14 @@ use std::env;
 fn main() {
     for (key, _) in env::vars() {
         let key = match key.as_str() {
-            "DEP_HDF5_HAVE_DIRECT" => "h5_have_direct".into(),
-            "DEP_HDF5_HAVE_STDBOOL" => "h5_have_stdbool".into(),
-            "DEP_HDF5_HAVE_PARALLEL" => "h5_have_parallel".into(),
-            "DEP_HDF5_HAVE_THREADSAFE" => "h5_have_threadsafe".into(),
-            "DEP_HDF5_MSVC_DLL_INDIRECTION" => "h5_dll_indirection".into(),
+            "DEP_HDF5_HAVE_DIRECT" => "have_direct".into(),
```
aldanor, line 6: Wonder, should we have `_` -> `-` and make all of these `feature=`?

src/hl/dataset.rs
```diff
@@ -907,8 +907,12 @@ macro_rules! impl_builder {
             self.builder.$name($($var),*); self
         }
     };
-    ($plist:ident: $name:ident($($var:ident: $ty:ty),*)) => {
+    ($(#[cfg(feature = $feature:literal)])*
```
aldanor, line 910: Just wondering, wouldn't something like `#[cfg($tt:tt)]` work, so you wouldn't have to hard-code specific syntax here?

Cargo.toml
mulimoen: Oh yeah, that is of course better, I assumed they would show up, but it seems not

aldanor:
(Was away for a little while, back now)

Wonder, for the docs.rs build, will we be able to get all the have-parallel, have-direct, etc entries show up? (and what can we do in order to make it work?)

src/hl/dataset.rs
mulimoen: I think I tried, but this was not as straight forward as I thought. My experience with macros is too limited to see what went wrong in this instance

hdf5-sys/src/h5s.rs
```diff
@@ -112,7 +115,7 @@ extern "C" {
     ) -> htri_t;
 }
 
-#[cfg(any(hdf5_1_12_0, hdf5_1_10_7))]
```
aldanor, line 115: @mulimoen Btw, what's this?... Shouldn't this be equivalent to just `1.10.7`?

mulimoen: Yes, that is equivalent. The function was backported so I kept `hdf5_1_12_0` for simplicity. It is superfluous now

aldanor:
@mulimoen Looks good to merge? (I'll probably update the changelog a bit)

We still need to figure how to build docs on:
- 1.12.1
- direct/parallel/threadsafe/mpio/blosc/lzf all enabled
(but that's a separate issue)

mulimoen:
Enabled this for `hdf5-sys` now. It's not ideal everywhere, some items such as `hbool_t` is suggested as only available with certain features enabled even though it is always present. This is very minor compared to seeing which functions are available in the current version. Looks good to merge now

PR #185, opened 2021-11-13
Fix multiple filter problems and add tests

### LINKED ISSUE #183: blosc_zstd() passes incorrect values to the filter
In my particular example, I'm using `.blosc_zstd(9, true)` in the dataset-with-data-builder and getting this in the output file without any errors:

```
    Storage:   409713920 logical bytes, 469757904 allocated bytes, 87.22% utilization
    Filter-0:  blosc-32001 OPT {2, 2, 4, 67108272, 0, 0, 0}
```

Note the `0, 0, 0` which should be `9, 1, 5`. What's interesting is that the first four values are correct.

Need  further investigation, e.g. minimal reproducible example.

aldanor:
Closes #183

### DIFF
--- a/src/hl/filters.rs
+++ b/src/hl/filters.rs
@@ -167,7 +167,7 @@ impl Filter {
     }
 
     pub fn get_info(filter_id: i32) -> FilterInfo {
-        if h5call!(H5Zfilter_avail(filter_id as _)).map(|x| x > 0).unwrap_or_default() {
+        if !h5call!(H5Zfilter_avail(filter_id as _)).map(|x| x > 0).unwrap_or_default() {
             return FilterInfo::default();
         }
         let mut flags: c_uint = 0;
--- a/src/hl/filters.rs
+++ b/src/hl/filters.rs
@@ -166,12 +166,12 @@ impl Filter {
         }
     }
 
-    pub fn get_info(filter_id: H5Z_filter_t) -> FilterInfo {
-        if h5call!(H5Zfilter_avail(filter_id)).map(|x| x > 0).unwrap_or_default() {
+    pub fn get_info(filter_id: i32) -> FilterInfo {
+        if h5call!(H5Zfilter_avail(filter_id as _)).map(|x| x > 0).unwrap_or_default() {
             return FilterInfo::default();
         }
         let mut flags: c_uint = 0;
-        h5lock!(H5Zget_filter_info(filter_id, &mut flags as *mut _));
+        h5lock!(H5Zget_filter_info(filter_id as _, &mut flags as *mut _));
         FilterInfo {
             is_available: true,
             encode_enabled: flags & H5Z_FILTER_CONFIG_ENCODE_ENABLED != 0,
--- a/src/hl/filters.rs
+++ b/src/hl/filters.rs
@@ -129,8 +129,8 @@ pub(crate) fn register_filters() {
     }
 }
 
-/// Returns `true` if gzip filter is available.
-pub fn gzip_available() -> bool {
+/// Returns `true` if deflate filter is available.
+pub fn deflate_available() -> bool {
     h5lock!(H5Zfilter_avail(H5Z_FILTER_DEFLATE) == 1)
 }
 
@@ -561,8 +561,8 @@ pub(crate) fn validate_filters(filters: &[Filter], type_class: H5T_class_t) -> R
 #[cfg(test)]
 mod tests {
     use super::{
-        blosc_available, gzip_available, lzf_available, szip_available, validate_filters, Filter,
-        SZip, ScaleOffset,
+        blosc_available, deflate_available, lzf_available, szip_available, validate_filters,
+        Filter, SZip, ScaleOffset,
     };
     use crate::test::with_tmp_file;
     use crate::{plist::DatasetCreate, Result};
@@ -571,7 +571,7 @@ mod tests {
     #[test]
     fn test_filter_pipeline() -> Result<()> {
         let mut comp_filters = vec![];
-        if gzip_available() {
+        if deflate_available() {
             comp_filters.push(Filter::deflate(3));
         }
         if szip_available() {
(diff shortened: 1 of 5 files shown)


aldanor:
This is WIP, but at least it's reproducible now.

aldanor:
As can be seen in the filter pipeline test, it's not just blosc/lzf filters, looks like something gets corrupted with the filter pipeline when applying it to the dataset.

aldanor:
@mulimoen Any thoughts where the corruption might happen? Hm..

mulimoen:
I am not very familiar with the filter code, but I've had a look at the `lzf` filter. In here the filter is modified in `set_local_lzf` which increases the number of elements to three which clashes with the check in `parse_lzf` which checks
```rust
ensure!(cdata.is_empty(), "expected length 0 cdata for lzf filter");
```
This check should probably be removed or check if 0 or 3?

mulimoen:
And I do wonder if `nbit` filter must also have a larger `cdata` to store `offset` and `precision`

aldanor:
> And I do wonder if `nbit` filter must also have a larger `cdata` to store `offset` and `precision`. The shuffle filter must store the number of bytes for shuffling.

Yep, agreed, and that logic is pretty complicated per-filter (i.e. cdata size may depend on the data itself), and in some cases there's up to around 20 values being stored. I've relaxed the checks so that only the minimum length is verified in the cases where we need to pull the parameters.

Now the failure you see in the filters test is the one where blosc filter parameters are zeroed out.

aldanor:
That is, before we apply the filter to the dataset:

```rust
// before applying to dataset (just storing in DCPL)
cdata = [
    0, 0, 0, 0, 9, 1, 5,
]
// after applying to dataset and then extracting it back
cdata = [
    2, 2, 4, 800000, 0, 0, 0,
]
```

Note that all internal parameters have been filled out now (first 4 values), but the user parameters (last 3) got erased somehow...

aldanor:
Was an actual bug - a single character mistake which essentially disabled all blosc filters 🤣 (hopefully fixed now in baf4ddb - and we'll have to release a version asap so that it actually works)

aldanor:
Interesting, now there's obscure conda-related failures.

aldanor:
Looks like the failures are related to `HDF5_PLUGIN_PATH` somehow

aldanor:
Any thoughts on how to reproduce/fix this? Hm...

aldanor:
Another thought that I had - don't hide `Blosc`, `BloscShuffle` structs etc under `#[feature = "blosc"]` so that you could read filter pipeline and see it properly (and not as `Filter::User`).

However, then we need to manually check that filters are available (both `cfg!(feature = "blosc")` and `blosc_available()`) when applying the filters to DCPL. Maybe it makes sense to do it regardless.

mulimoen:
I tried reproducing in python here: https://github.com/mulimoen/hdf5_py_test/actions/runs/1473774852 but could not get the error to trigger. Maybe `h5py` has a workaround?

mulimoen:
> Another thought that I had - don't hide Blosc, BloscShuffle structs etc under #[feature = "blosc"] so that you could read filter pipeline and see it properly (and not as Filter::User).

Would this not depend on `BLOSC_FILTER_ID` being exactly `32001`? This might be a problem for all files created by other libraries. (This is also a breaking change)

aldanor:
> > Another thought that I had - don't hide Blosc, BloscShuffle structs etc under #[feature = "blosc"] so that you could read filter pipeline and see it properly (and not as Filter::User).
> 
> Would this not depend on `BLOSC_FILTER_ID` being exactly `32001`? This might be a problem for all files created by other libraries. (This is also a breaking change)

Blosc filter ID being 32001 is a sort of immutable invariant that we have to accept, I guess... More details here: https://support.hdfgroup.org/pubs/rfcs/RFC_Filters-registration-policy-final.pdf

aldanor:
Ok, I was able to reproduce it, same as one of the failures in CI: macOS, using 1.8.18 from anaconda main channel.

```
Error: H5Pset_filter(): failed to call private function: can't open directory: .../envs/hdf5-1.8.18/lib/hdf5/plugin
```

Apparently, there's `H5PL*` API for managing plugin paths etc, but it's only available on 1.10.1 and later. Before that it just looks at env vars which may be cached or not... I'll give it a try tomorrow to see if anything can be hacked around.

aldanor:
Hm, if setting `HDF5_PLUGIN_PATH` manually to some obscure folder, it now yields

```
Error: H5Pset_filter(): failed to call private function: failed to load dynamically loaded plugin
```

So just setting the env var wouldn't solve it, need to dig deeper.

aldanor:
Ok, interesting - it's adding that weird non-existent user filter that fails *sometimes* in some conditions

If I remove a line with `Filter::user(...)`, it seems to pass. Wonder then, why doesn't it fail everywhere, hm.

aldanor:
Ok, fixed yet another typo/bug in a573e72 (🤦, having proper tests is a good thing, we need more of them to cover all the new dataset/DCPL stuff...), let's see how this works out now.

aldanor:
Ok, this should be good to go, I think. Hopefully fixes most filter problems.

src/hl/filters.rs
```diff
@@ -121,16 +123,16 @@ pub struct FilterInfo {
...
     if let Err(e) = blosc::register_blosc() {
-        eprintln!("{}", e);
+        eprintln!("Error while registering Blosc filter: {}", e);
     }
 }
 
-/// Returns `true` if gzip filter is available.
-pub fn gzip_available() -> bool {
+/// Returns `true` if deflate filter is available.
+pub fn deflate_available() -> bool {
```
mulimoen, line 135: Add `gzip_available` but mark as deprecated to keep semver to minor

mulimoen (changes requested):
Looks good apart from the breaking change

src/hl/filters.rs
aldanor: Yea, good point

mulimoen (approved):

PR #188, opened 2021-11-19
Add support for `#[repr(packed)]` and `#[repr(transparent)]` in hdf5-derive

### LINKED ISSUE #187: Support `repr(packed)` and `repr(transparent)` in `#[derive(H5Type)]`
- `#[repr(packed)]` should be simply allowed, it will use field offsets from a packed struct
- `#[repr(transparent)]` for a single-field structs should make the struct type equivalent to field type

aldanor:
Closes #187.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - `Error` now implements `From<Infallible>`, which allows passing convertible
   extents (like tuples of integers) where `impl TryInto<Extents>` is required.
 - Support for HDF5 versions 1.12.1 and 1.10.8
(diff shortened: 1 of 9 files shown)


mulimoen:
Is `repr(packed)` abi safe without `repr(C)`?

aldanor:
> Is `repr(packed)` abi safe without `repr(C)`?

I think it's not our business - i.e. what the users are doing with their packed structs in Rust. There's tons of ways that come back to bite you, of course (unaligned loads etc).

What's important, I think, is that packed/C/transparent repr of a struct will be identical on any platform?

// One other reason why packed repr is important btw and a sort of a standard is that in numpy everything is packed by default.

mulimoen (approved):

PR #190, opened 2021-11-24
Simplify feature expansion in macro

mulimoen:
It turns out there is a fragment specifier for features: https://veykril.github.io/tlborm/decl-macros/minutiae/fragment-specifiers.html#meta

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


aldanor:
Ah, yes, the `meta` specifier, indeed! Looks good, thanks :)

aldanor (approved):

aldanor:
(maybe update the changelog though?)

PR #191, opened 2021-12-03
Fix latest round of clippy lints

mulimoen:


hdf5-sys/build.rs
```diff
@@ -111,17 +116,20 @@ fn validate_runtime_version(config: &Config) {
...
-            println!("Adding extra link paths (ld)...");
-            for caps in re.captures_iter(&ldv) {
-                let path = &caps["path"];
-                println!("    {}", path);
-                link_paths.push(path.into());
-            }
-        } else {
-            println!("Unable to add extra link paths (ld).");
-        }
+        run_command("ld", &["--verbose"]).map_or_else(
```
aldanor, line 119: Tbh, these `.map_or_else()` actually make it much less readable :( Would clippy be against a simple match as well? Maybe we can disable this particular lint?

hdf5-sys/build.rs
```diff
@@ -221,6 +231,7 @@ fn get_conf_header<P: AsRef<Path>>(inc_dir: P) -> PathBuf {
     }
 }
 
+#[must_use]
```
aldanor, line 234: `#[must_use]` is very questionable in our case, and clippy would recommend every single thing to be marked with it - I'd say add a lint ignore for it and remove must-uses?

hdf5-types/src/dyn_value.rs
```diff
@@ -527,12 +534,14 @@ unsafe impl DynClone for DynVarLenString<'_> {
...
+            debug_assert!(is_aligned_for_type::<DynVarLenString>(out.as_ptr()));
             unsafe {
                 let raw_len = self.raw_len();
-                let dst = crate::malloc(raw_len + 1) as *mut _;
+                let dst = crate::malloc(raw_len + 1).cast();
                 ptr::copy_nonoverlapping(self.get_ptr(), dst, raw_len);
-                *dst.add(raw_len) = 0;
-                *(out.as_mut_ptr() as *mut *const u8) = dst as _;
+                dst.add(raw_len).write(0);
+                let outptr = out.as_mut_ptr().cast::<*const u8>(); // alignment asserted above
```
aldanor, line 543: Hm, not sure about this. How can a `*const u8` be non-aligned?

What's the story about the align checks in this pr?

src/hl/dataset.rs
```diff
@@ -456,13 +457,14 @@ impl DatasetBuilderInner {
...
         let chunking_allowed = extents.size() > 0 || extents.is_resizable();
 
-        let chunk = if let Some(chunk) = &self.chunk {
-            chunk.clone()
-        } else if chunking_required && chunking_allowed {
-            Chunk::MinKB(DEFAULT_CHUNK_SIZE_KB)
-        } else {
-            Chunk::None
-        };
+        let chunk = self.chunk.as_ref().map_or(
```
aldanor, line 460: Same as with `.map_or_else()` case above, not sure I agree this is more readable? (I'd probably ignore it along with map-or-else)

aldanor:
@mulimoen Thanks for taking care of all this - many lints actually make sense - e.g. there was some old code that couldn't use `[T; N]` deriving Clone for all `N` so we had to do it manually, etc.

There's a few things I've pointed out that are questionable:
- must-use
- map-or, map-or-else that, in my opinion, hurt readability in our particular use cases
- what's with align checks, which lint is it? (especially when casting to byte pointers)

hdf5-types/src/dyn_value.rs
mulimoen: This pointer is an `*mut *const u8` which needs 4/8-bytes of alignment

hdf5-sys/build.rs
mulimoen: Blacklisted and removed `#[must_use]`

src/hl/dataset.rs
mulimoen: Blacklisted and reverted this code

mulimoen:
Thanks @aldanor for reviewing this. Align checks are for casting to pointer-to-pointer, e.g. error: casting from `*mut u8` to a more-strictly-aligned pointer (`*mut *const u8`) (1 < 8 bytes). Could be I have misunderstood some of this code, quite a lot of tricky pointer logic there.

aldanor:
Given that we allocate memory either via `libc::malloc`, whose [man](http://kernel.org/doc/man-pages/online/pages/man3/malloc.3.html) states:

> The `malloc()` and `calloc()` functions return a pointer to the allocated memory that is suitably aligned for **any** kind of variable.

Or via `H5allocate_memory`:
- which calls `H5MM_malloc`
  - which either calls `HDmalloc`
    - which is defined as alias for `malloc` => all good, no checks needed
  - or uses an own linked-list allocator
    - which allocates an `H5MM_block_t` block via `HDmalloc` (so blocks are 8-byte aligned at the very least)
    - each block has a header of `H5MM_HEAD_GUARD_SIZE = 8`, so each block start is 8-byte aligned
    - inside the block itself, buffer is also 8-byte aligned as there's a specific dummy field inside a union to ensure that:
      ```c
      double _align; /* Align following buffer (b) to double boundary (unused) */
      ```
    - so, buffer start is 8-byte aligned at the very least

=> in whichever case, 8-byte alignment is pretty much guaranteed and we don't need to check it ourselves?

mulimoen:
That is some great detective work @aldanor! I would also not expect to see any alignment issues as allocations in `C` tends to be very conservative, so it's great that we can remove these checks

src/hl/container.rs
```diff
@@ -26,12 +26,14 @@ impl<'a> Reader<'a> {
     }
 
     /// Set maximum allowed conversion level.
+    #[must_use]
```
aldanor, line 29: Maybe remove the new `#[must_use]` instances in this crate as well?

hdf5-derive/src/lib.rs
```diff
@@ -176,9 +175,8 @@ fn impl_trait(
                     .filter(|&(_, f)| !is_phantom_data(&f.ty))
                     .map(|(i, f)| (Index::from(i), f))
                     .unzip();
-                if fields.is_empty() {
-                    panic!("Cannot derive H5Type for empty tuple structs");
-                }
+                assert!(!fields.is_empty(), "Cannot derive H5Type for empty tuple structs");
```
aldanor, line 178: Todo note to self: we should really use `#[proc_macro_error]` which will emit nice errors attached to spans (it's just when hdf5-derive was written, proc-macro-error wasn't there).

src/hl/container.rs
mulimoen: `#[must_use]` has been removed from everywhere I've added them

hdf5-sys/build.rs
```diff
@@ -1,3 +1,6 @@
+#![allow(clippy::must_use_candidate)]
+#![allow(clippy::option_if_let_else)]
+use std::convert::TryInto;
```
aldanor, line 3: (cosmetic, for consistency sake)

```suggestion
#![allow(clippy::option_if_let_else)]

use std::convert::TryInto;
```

aldanor (approved):

aldanor:
Great, thanks!

PR #192, opened 2022-01-04
Support hdf5 version 1.13

mulimoen:
This PR adds 1.13 as a supported version, but does not use any of the new functionality.

The list of changes are found here: https://confluence.hdfgroup.org/display/support/HDF5%201.13.0#compatibility1130

### DIFF
--- a/hdf5-src/Cargo.toml
+++ b/hdf5-src/Cargo.toml
@@ -32,7 +32,6 @@ hl = []
 zlib = ["libz-sys"]
 deprecated = []
 threadsafe = []
-default = ["deprecated"]
 
 [dependencies]
 libz-sys = { version = "1.0.25", features = ["static"], optional = true, default-features=false }
--- a/hdf5-src/Cargo.toml
+++ b/hdf5-src/Cargo.toml
@@ -32,6 +32,7 @@ hl = []
 zlib = ["libz-sys"]
 deprecated = []
 threadsafe = []
+default = ["deprecated"]
 
 [dependencies]
 libz-sys = { version = "1.0.25", features = ["static"], optional = true, default-features=false }
--- a/hdf5-sys/src/h5es.rs
+++ b/hdf5-sys/src/h5es.rs
@@ -2,6 +2,7 @@
 use crate::internal_prelude::*;
 
 #[repr(C)]
+#[derive(Debug)]
 pub struct H5ES_err_info_t {
     api_cname: *mut c_char,
     api_args: *mut c_char,
@@ -19,6 +20,7 @@ pub struct H5ES_err_info_t {
 }
 
 #[repr(C)]
+#[derive(Debug)]
 pub struct H5ES_op_info_t {
     api_cname: *const c_char,
     api_args: *mut c_char,
@@ -34,6 +36,7 @@ pub struct H5ES_op_info_t {
 }
 
 #[repr(C)]
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
 pub enum H5ES_status_t {
     H5ES_STATUS_IN_PROGRESS,
     H5ES_STATUS_SUCCEED,
--- a/hdf5-sys/src/h5es.rs
+++ b/hdf5-sys/src/h5es.rs
@@ -1,0 +1,84 @@
+//! Event set module
+use crate::internal_prelude::*;
+
+#[repr(C)]
+pub struct H5ES_err_info_t {
+    api_cname: *mut c_char,
+    api_args: *mut c_char,
+
+    app_file_name: *mut c_char,
+    app_func_name: *mut c_char,
+    app_line_num: c_uint,
+
+    op_ins_count: u64,
+    op_ins_ts: u64,
+    op_exec_ts: u64,
+    op_exec_time: u64,
+
+    err_stack_id: hid_t,
+}
+
+#[repr(C)]
+pub struct H5ES_op_info_t {
+    api_cname: *const c_char,
+    api_args: *mut c_char,
+
+    app_file_name: *const c_char,
+    app_func_name: *const c_char,
+    app_line_num: c_uint,
+
+    op_ins_count: u64,
+    op_ins_ts: u64,
+    op_exec_ts: u64,
+    op_exec_time: u64,
+}
+
+#[repr(C)]
+pub enum H5ES_status_t {
+    H5ES_STATUS_IN_PROGRESS,
+    H5ES_STATUS_SUCCEED,
+    H5ES_STATUS_CANCELED,
+    H5ES_STATUS_FAIL,
+}
+
+pub type H5ES_event_complete_func_t = Option<
+    extern "C" fn(
+        op_info: *mut H5ES_op_info_t,
+        status: H5ES_status_t,
+        err_stack: hid_t,
+        ctx: *mut c_void,
+    ) -> c_int,
+>;
+
+pub type H5ES_event_insert_func_t =
+    Option<extern "C" fn(op_info: *const H5ES_op_info_t, ctx: *mut c_void) -> c_int>;
+
+extern "C" {
+    pub fn H5ESinsert_request(es_id: hid_t, connector_id: hid_t, request: *mut c_void) -> herr_t;
+}
+
+extern "C" {
+    pub fn H5EScancel(
+        es_id: hid_t, num_not_canceled: *mut size_t, err_occured: *mut hbool_t,
+    ) -> herr_t;
+    pub fn H5ESclose(es_id: hid_t) -> herr_t;
+    pub fn H5EScreate() -> hid_t;
+    pub fn H5ESfree_err_info(num_err_info: size_t, err_info: *mut H5ES_err_info_t) -> herr_t;
+    pub fn H5ESget_count(es_id: hid_t, count: *mut size_t) -> herr_t;
+    pub fn H5ESget_err_count(es_id: hid_t, num_errs: *mut size_t) -> herr_t;
+    pub fn H5ESget_err_info(
+        es_id: hid_t, num_err_info: size_t, err_info: *mut H5ES_err_info_t,
+        err_cleared: *mut size_t,
+    ) -> herr_t;
+    pub fn H5ESget_err_status(es_id: hid_t, err_occured: *mut hbool_t) -> herr_t;
+    pub fn H5ESget_op_counter(es_id: hid_t, counter: *mut u64) -> herr_t;
+    pub fn H5ESregister_complete_func(
+        es_id: hid_t, func: H5ES_event_complete_func_t, ctx: *mut c_void,
+    ) -> herr_t;
+    pub fn H5ESregister_insert_func(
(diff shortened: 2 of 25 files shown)


hdf5-src/Cargo.toml
```diff
@@ -30,6 +30,7 @@ hl = []
 zlib = ["libz-sys"]
 deprecated = []
 threadsafe = []
+default = ["deprecated"]
```
aldanor, line 33: Wondering why add this?

hdf5-sys/src/h5es.rs
```diff
@@ -0,0 +1,84 @@
+//! Event set module
+use crate::internal_prelude::*;
+
+#[repr(C)]
+pub struct H5ES_err_info_t {
```
aldanor, line 5: Add common derives (like in other modules) for all these structs?

hdf5-sys/src/h5es.rs
```diff
@@ -0,0 +1,84 @@
...
+    app_line_num: c_uint,
+
+    op_ins_count: u64,
+    op_ins_ts: u64,
+    op_exec_ts: u64,
+    op_exec_time: u64,
+}
+
+#[repr(C)]
+pub enum H5ES_status_t {
```
aldanor, line 37: For enums:
- Common derives?
- Specify values explicitly?

hdf5-sys/src/h5vl.rs
```diff
@@ -1,12 +1,1530 @@
...
+    h5o::{H5O_info2_t, H5O_iterate2_t, H5O_token_t, H5O_type_t},
+};
 
 pub type H5VL_class_value_t = c_int;
 
 // Incomplete type
+#[cfg(not(feature = "1.13.0"))]
 pub type H5VL_class_t = c_void;
 
+#[cfg(feature = "1.13.0")]
```
aldanor, line 24: Since there's such a huge 1.13 chunk, maybe pack it all into one private submodule and re-export?

E.g.

```rust
#[cfg(feature = "1.13.0")]
mod v1_13_0 {
    pub struct H5VL_info_class_t {}
}

#[cfg(feature = "1.13.0")]
pub use self::v_1_13_0::*;

**Comment by aldanor** [comment]
This is a pretty humongous diff, thanks for taking on it. I left a few comments if you don't mind, and maybe squash it all before merging then?

(I was also wondering if we're doing it right with the regex stuff in the build script... but that's a separate question)

File: hdf5-src/Cargo.toml (hunk 1 above)
Comment (mulimoen) [reply hdf5-src/Cargo.toml:33]: In `src/hl/dataspace.rs` we are using `H5Sencode1` which is not included when building `hdf5-1.13` from source

File: hdf5-sys/src/h5es.rs (hunk 2 above)
Comment (mulimoen) [reply hdf5-sys/src/h5es.rs:6]: Added derives

File: hdf5-sys/src/h5es.rs (hunk 3 above)
Comment (mulimoen) [reply hdf5-sys/src/h5es.rs:40]: Derives added, but values are not specified since they are not specified in `hdf5-c`

File: hdf5-sys/src/h5vl.rs (hunk 4 above)
Comment (mulimoen) [reply hdf5-sys/src/h5vl.rs:24]: I'm not sure if this is the best approach as a lot of these functions are not introduced in 1.13, but should be backported to 1.12 whilst ensuring breaking changes are accounted for

File: hdf5-sys/src/h5es.rs (hunk 3 above)
Comment (aldanor) [reply hdf5-sys/src/h5es.rs:40]: Fair enough, my bad, I guess I haven't dealt with C enums in a long while; just checked - the generated size by default is 4 (i.e. same as `c_int`) and the first value is 0, so it will try to mimic C behaviour on the platform.

File: hdf5-src/Cargo.toml (hunk 1 above)
Comment (aldanor) [reply hdf5-src/Cargo.toml:33]: Hm, shouldn't we rather have a `cfg_if` there so that `H5Sencode1` is only used pre-1.13?

File: hdf5-sys/src/h5vl.rs (hunk 4 above)
Comment (aldanor) [reply hdf5-sys/src/h5vl.rs]: So maybe it could be split between `mod v1_12_0` and `mod v1_13_0` then? It's just that there's a ton of these new functions and definitions here

**Comment by aldanor** [comment]
This should be probably rebased on the latest master with clippy lints

File: hdf5-sys/src/h5vl.rs (hunk 4 above)
Comment (mulimoen) [reply hdf5-sys/src/h5vl.rs]: Moved all new functions to a separate module

File: hdf5-sys/src/h5es.rs [hunk 5]
```diff
@@ -0,0 +1,87 @@
+//! Event set module
+use crate::internal_prelude::*;
+
+#[repr(C)]
+#[derive(Debug)]
```
aldanor, line 5: All the structs should probably be `Copy / Clone` (even if they contain pointers) because they are such in C

hdf5-sys/src/h5fd.rs
```diff
@@ -207,6 +214,16 @@ pub struct H5FD_class_t {
     >,
     pub unlock:
         Option<extern "C" fn(file: *mut H5FD_t, oid: *mut c_uchar, last: hbool_t) -> herr_t>,
+    pub del: Option<extern "C" fn(name: *const c_char, fapl: hid_t) -> herr_t>,
```
aldanor, line 217: Shouldn't there be `#[cfg(1.13)]` here?

hdf5-sys/src/h5vl.rs
```diff
@@ -1,12 +1,64 @@
...
+};
 
 pub type H5VL_class_value_t = c_int;
 
 // Incomplete type
+#[cfg(all(feature = "1.12.0", not(feature = "1.13.0")))]
 pub type H5VL_class_t = c_void;
 
+#[repr(C)]
+#[derive(Debug)]
```
aldanor, line 13: `Copy/Clone` for all structs here as well?

hdf5-sys/src/h5vl.rs
```diff
@@ -48,3 +100,1905 @@ extern "C" {
...
+    #[repr(C)]
+    #[derive(Debug)]
+    pub struct H5VL_attr_get_info_args_t {
+        pub loc_params: H5VL_loc_params_t,
+        pub attr_name: *const c_char,
+        pub ainfo: *mut H5A_info_t,
+    }
+
+    #[repr(C)]
+    pub union H5VL_attr_get_args_t_union {
```
aldanor, line 198: Can `Copy/Clone` be derived for unions?

hdf5-sys/src/h5vl.rs
```diff
@@ -48,3 +88,1918 @@ extern "C" {
...
+        pub get_type: ManuallyDrop<H5VL_attr_get_args_t_union_get_type>,
+    }
+
+    #[repr(C)]
+    pub struct H5VL_attr_get_args_t {
+        pub op_type: H5VL_attr_get_t,
+        pub args: H5VL_attr_get_args_t_union,
+    }
+
+    impl Debug for H5VL_attr_get_args_t {
```
aldanor, line 213: Maybe instead just `impl Debug` for the union? And then you can simply derive Debug for this struct

mulimoen: `impl Debug` allows the use of the tag to decipher the union, but could could prefer a dummy impl `f.write("<UNION>")` instead

aldanor: Ahh yea, I'm being dumb, the union is untagged so we can't format it. Sorry, I'm probably just too used to Rust enums 😄 

All good then.

hdf5-sys/src/h5vl.rs
mulimoen: Yes, they are treated as just bytes

hdf5-sys/src/h5fd.rs
mulimoen: Well spotted!

mulimoen:
I did want to remove some of the boilerplate in `hv5l.rs` with
```rust
macro_rules! impl_debug_args {
    ($struct: ty, $($m: pat => $arg: expr),+) => {
        impl Debug for $struct {
            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
                let s = f.debug_struct(stringify!($struct)).field("op_type", &self.op_type);
                let args = &self.args;
                let s = match self.op_type {
                    $($m => s.field("args", $arg)),+,
                };
                s.finish()
            }
        }
    };
}

impl_debug_args!(H5VL_link_specific_args_t,
    H5VL_link_specific_t::H5VL_LINK_DELETE => "delete",
    H5VL_link_specific_t::H5VL_LINK_EXISTS => self.args.exists,
    H5VL_link_specific_t::H5VL_LINK_ITER => self.args.iterate
);
```
But one can't use `self` in a macro

aldanor:
Hm, I don't think adding a dependency (`paste`), especially with proc macros, is worth it to simplify a few lines of code.

I think you can do this instead:

```rust
macro_rules! impl_debug_args {
    ($ty:ident, $tag:ident, $args:ty, {$($variant:tt => $func:expr),+$(,)*}) => {
        impl std::fmt::Debug for $ty {
            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
                let mut s = f.debug_struct(stringify!($ty));
                s.field("op_type", &self.op_type);
                match self.op_type {$(
                    $tag::$variant => {
                        s.field("args", &($func as fn($args) -> _)(self.args));
                    },
                )+}
                s.finish()
            }
        }
    };
}
```

which can be then used as

```rust
impl_debug_args!(
    H5VL_link_specific_args_t,
    H5VL_link_specific_t,
    H5VL_link_specific_args_t_union,
    {
        H5VL_LINK_DELETE => |_| "delete",
        H5VL_LINK_EXISTS => |args| unsafe { args.exists },
        H5VL_LINK_ITER => |args| unsafe { args.iterate },
    }
);
```

aldanor:
Or, even better, `"delete"` is not needed at all here. So you can do:

```rust
macro_rules! impl_debug_args {
    ($ty:ident, $tag:ident, $args:ty, {$($variant:tt => $func:expr),+$(,)*}) => {
        #[allow(unreachable_patterns)]
        impl std::fmt::Debug for $ty {
            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
                let mut s = f.debug_struct(stringify!($ty));
                s.field("op_type", &self.op_type);
                match self.op_type {$(
                    $tag::$variant => {
                        s.field("args", &($func as fn($args) -> _)(self.args));
                    }
                    _ => {}
                )+}
                s.finish()
            }
        }
    };
}

impl_debug_args!(
    H5VL_link_specific_args_t,
    H5VL_link_specific_t,
    H5VL_link_specific_args_t_union,
    {
        H5VL_LINK_EXISTS => |args| unsafe { args.exists },
        H5VL_LINK_ITER => |args| unsafe { args.iterate },
    }
);
```

mulimoen:
The macro looks quite good right now, thanks for the help @aldanor!

aldanor:
Looks good now, I think! Thanks for the big PR, should be ready to go.

aldanor (approved):

PR #196, opened 2022-01-27
Use proc-macro-error

mulimoen:
As mentioned in #191 it would improve on the error messages to include the span. This highlights the struct/enum instead of giving the general `proc_macro panicked` and the message in the help section.

### DIFF
--- a/hdf5-derive/src/lib.rs
+++ b/hdf5-derive/src/lib.rs
@@ -1,12 +1,11 @@
 #![recursion_limit = "192"]
 
-extern crate proc_macro;
-
 use std::iter;
 use std::mem;
 use std::str::FromStr;
 
 use proc_macro2::{Ident, Span, TokenStream};
+use proc_macro_error::{abort, proc_macro_error};
 use quote::{quote, ToTokens};
 use syn::{
     parse_macro_input, AttrStyle, Attribute, Data, DeriveInput, Expr, Fields, Index, Meta,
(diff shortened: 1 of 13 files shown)


hdf5-derive/tests/compile-fail/empty-struct.stderr
```diff
@@ -1,7 +1,5 @@
-error: proc-macro derive panicked
- --> $DIR/empty-struct.rs:4:10
+error: Cannot derive `H5Type` for empty structs
```
aldanor, line 1: minor, for consistency sake, can we make all error messages (that don't start with a `) start with lowercase letter?

e.g. `cannot derive ...`

hdf5-derive/src/lib.rs
```diff
@@ -1,19 +1,22 @@
 #![recursion_limit = "192"]
 
 extern crate proc_macro;
```
aldanor, line 3: Hm, do we even need these `extern crate` anymore?

aldanor:
Very nice, this looks like proper compiler error messages now

hdf5-derive/src/lib.rs
mulimoen: Let's see if this works with our `MSRV`

hdf5-derive/tests/compile-fail/empty-struct.stderr
mulimoen: Fixed

hdf5-derive/src/lib.rs
mulimoen: This works but a CI job is halting (most likely unrelated)

aldanor: Unrelated, I think, otherwise it'd crash immediately

aldanor (approved):

PR #197, opened 2022-01-28
Move catch-all in match after variants

mulimoen:
The previous macro would always match on the second item even if more than one variant was present


aldanor:
Indeed :) Thanks, good catch!

aldanor (approved):

PR #198, opened 2022-01-31
Add ByteReader implementing io::Read and io::Seek

### LINKED ISSUE #193: Implement io::Read + io::Seek for hdf5::Reader
I have some encoded data stored within a hdf5::Dataset. Currently I'm using:

```rust
let ds = self.group()?.dataset("some_key")?;
let encoded = ds.read_1d()?;
let mut reader = SomeDecoderThatNeedsIORead::new(Cursor::new(encoded.as_slice().unwrap()))?;
```
or if I only want to read a meta data header:
```rust
let encoded = ds.read_slice_1d(s![..1024]) // I need to know the header-size before hand. This may vary.
let mut reader = SomeDecoderThatNeedsIORead::new(Cursor::new(encoded.as_slice().unwrap()))?;
```

It would be nice to just do:
```rust
let mut reader = SomeDecoderThatNeedsIORead::new(ds);
```

I guess `std::io::Read` should be simpler to implement, but `Seek` would also be nice.

In the first place, an uncompressed dataset (i.e. no gzip etc) would be sufficient since compression does not bring any benefit for already encoded data. Not sure if this makes a difference.

Rikorose:
The design is based on io::Cursor. This PR only addresses datasets. I don't think it makes sense to also support io::Read for attributes since the whole advantage of io::Read is to only read the relevant part and not all bytes at once.

Currently I get the following error during H5DRead:
```
Error: IoError(Custom { kind: Other, error: "H5Dread(): can't read data: not a datatype" })
```
Closes #193

### DIFF
--- a/src/hl/container.rs
+++ b/src/hl/container.rs
@@ -1,5 +1,6 @@
 use std::convert::TryInto;
 use std::fmt::{self, Debug};
+use std::io;
 use std::mem;
 use std::ops::Deref;
 
@@ -346,6 +347,115 @@ impl<'a> Writer<'a> {
     }
 }
 
+#[derive(Debug)]
+pub struct ByteReader<'a> {
+    obj: &'a Container,
+    pos: u64,
+    dt: Datatype,
+    obj_space: Dataspace,
+    xfer: PropertyList,
+}
+
+impl<'a> ByteReader<'a> {
+    pub fn new(obj: &'a Container) -> Result<Self> {
+        ensure!(!obj.is_attr(), "ByteReader cannot be used on attribute datasets");
+
+        let file_dtype = obj.dtype()?;
+        let mem_dtype = Datatype::from_type::<u8>()?;
+        file_dtype.ensure_convertible(&mem_dtype, Conversion::NoOp)?;
+
+        let obj_space = obj.space()?;
+        ensure!(obj_space.shape().len() == 1, "Only rank 1 datasets can be read vie ByteReader.");
+        let xfer = PropertyList::from_id(h5call!(H5Pcreate(*crate::globals::H5P_DATASET_XFER))?)?;
+        if !hdf5_types::USING_H5_ALLOCATOR {
+            crate::hl::plist::set_vlen_manager_libc(xfer.id())?;
+        }
+        Ok(ByteReader { obj, pos: 0, obj_space, dt: mem_dtype, xfer })
+    }
+
+    pub fn set_position(&mut self, pos: u64) {
+        self.pos = pos;
+    }
+
+    fn dataset_len(&self) -> usize {
+        self.obj_space.shape()[0]
+    }
+
+    fn remaining_len(&self) -> usize {
+        self.dataset_len().saturating_sub(self.pos as usize)
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.pos >= self.dataset_len() as u64
+    }
+}
+
+trait IOResultExt<T> {
+    fn to_io_err(self) -> std::io::Result<T>;
+}
+impl<T> IOResultExt<T> for Result<T> {
+    fn to_io_err(self) -> std::io::Result<T> {
+        match self {
+            Ok(x) => Ok(x),
+            Err(e) => Err(std::io::Error::new(std::io::ErrorKind::Other, e.to_string())),
+        }
+    }
+}
+
+impl<'a> std::io::Read for ByteReader<'a> {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        let pos = self.pos as usize;
+        let amt = std::cmp::min(buf.len(), self.remaining_len());
+        let selection = Selection::new(pos..pos + amt);
+        let out_shape = selection.out_shape(&self.obj_space.shape()).to_io_err()?;
+        let fspace = self.obj_space.select(selection).to_io_err()?;
+        let mspace = Dataspace::try_new(&out_shape).to_io_err()?;
+        h5call!(H5Dread(
+            self.obj.id(),
+            self.dt.id(),
+            mspace.id(),
+            fspace.id(),
+            self.xfer.id(),
+            buf.as_mut_ptr().cast()
+        ))
+        .to_io_err()?;
+        self.pos += amt as u64;
+        Ok(out_shape[0])
+    }
+}
+
+impl<'a> std::io::Seek for ByteReader<'a> {
+    fn seek(&mut self, style: std::io::SeekFrom) -> std::io::Result<u64> {
+        let (base_pos, offset) = match style {
+            std::io::SeekFrom::Start(n) => {
+                self.pos = n;
+                return Ok(n);
+            }
+            std::io::SeekFrom::End(n) => (self.dataset_len() as u64, n),
+            std::io::SeekFrom::Current(n) => (self.pos, n),
+        };
+        let new_pos = if offset.is_negative() {
+            base_pos.checked_sub(offset.wrapping_abs() as u64)
+        } else {
+            base_pos.checked_add(offset as u64)
+        };
+        match new_pos {
+            Some(n) => {
+                self.pos = n;
+                Ok(self.pos)
+            }
+            None => Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "invalid seek to a negative or overflowing position",
+            )),
+        }
+    }
+
+    fn stream_position(&mut self) -> io::Result<u64> {
+        Ok(self.pos)
+    }
+}
+
 #[repr(transparent)]
 #[derive(Clone)]
 /// An object which can be read or written to.
@@ -393,6 +503,11 @@ impl Container {
     /// set custom type conversion options when writing.
     pub fn as_writer(&self) -> Writer {
         Writer::new(self)
+    }
+
+    /// Convert into a ``ByteReader`` that implements ``std::io::{Read, Seek}``.
+    pub fn as_byte_reader(&self) -> Result<ByteReader> {
+        ByteReader::new(self)
     }
 
     /// Returns the datatype of the dataset/attribute.
--- a/src/hl/container.rs
+++ b/src/hl/container.rs
@@ -505,7 +505,9 @@ impl Container {
         Writer::new(self)
     }
 
-    /// Convert into a ``ByteReader`` that implements ``std::io::{Read, Seek}``.
+    /// Creates a ``ByteReader`` which implements ``std::io::{Read, Seek}``.
+    /// ``ByteReader`` only supports 1D ``u8`` datasets. Attributes or datasets
+    /// with other dimensions or datatypes cannot be read with ``ByteReader``.
     pub fn as_byte_reader(&self) -> Result<ByteReader> {
         ByteReader::new(self)
     }
--- a/src/hl/container.rs
+++ b/src/hl/container.rs
@@ -505,9 +505,10 @@ impl Container {
         Writer::new(self)
     }
 
-    /// Creates a ``ByteReader`` which implements ``std::io::{Read, Seek}``.
-    /// ``ByteReader`` only supports 1D ``u8`` datasets. Attributes or datasets
-    /// with other dimensions or datatypes cannot be read with ``ByteReader``.
+    /// Creates `ByteReader` which implements [`Read`](std::io::Read)
+    /// and [`Seek`](std::io::Seek).
+    ///
+    /// ``ByteReader`` only supports 1-D `u8` datasets.
     pub fn as_byte_reader(&self) -> Result<ByteReader> {
         ByteReader::new(self)
     }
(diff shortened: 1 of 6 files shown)


Rikorose:
The CmpU8 trait is rather ugly, I just wanted to reuse the `test_read()` stuff. I guess it makes sense to add this as a new test.

However, it demonstrates the the current error: `H5Dread(): can't read data: not a datatype`

src/hl/container.rs
```diff
@@ -346,6 +347,109 @@ impl<'a> Writer<'a> {
     }
 }
```
mulimoen, line 349:
```suggestion
#[derive(Debug)]
```

src/hl/container.rs
```diff
@@ -346,6 +347,109 @@ impl<'a> Writer<'a> {
...
 
+pub struct ByteReader<'a> {
+    obj: &'a Container,
+    pos: u64,
+    tp_id: hid_t,
```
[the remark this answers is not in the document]
mulimoen: This is also why your test is failing, you are freeing the datatype before storing it in `ByteReader`

src/hl/container.rs
```diff
@@ -346,6 +347,109 @@ impl<'a> Writer<'a> {
...
+            Ok(x) => Ok(x),
+            Err(e) => Err(std::io::Error::new(std::io::ErrorKind::Other, e.to_string())),
+        }
+    }
+}
+
+impl<'a> std::io::Read for ByteReader<'a> {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        let pos = self.pos as usize;
+        let selection = Selection::new(pos..pos + buf.len());
```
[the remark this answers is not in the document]
Rikorose: Yes I guess it makes sense. io::Read states that this is implementation specific, and allowed to go beyond the overall length. But I cannot think of a real use case, so I will upward limit it.

src/hl/container.rs
```diff
@@ -346,6 +347,115 @@ impl<'a> Writer<'a> {
...
+    fn to_io_err(self) -> std::io::Result<T> {
+        match self {
+            Ok(x) => Ok(x),
+            Err(e) => Err(std::io::Error::new(std::io::ErrorKind::Other, e.to_string())),
+        }
+    }
+}
+
+impl<'a> std::io::Read for ByteReader<'a> {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
```
mulimoen, line 406: Not sure if it is a good idea, but this could be written
```rust
        || -> Result<usize> {
            let pos = self.pos as usize;
            let amt = std::cmp::min(buf.len(), self.remaining_len());
            let selection = Selection::new(pos..pos + amt);
            let out_shape = selection.out_shape(&self.obj_space.shape())?;
            let fspace = self.obj_space.select(selection)?;
            let mspace = Dataspace::try_new(&out_shape)?;
            h5call!(H5Dread(
                self.obj.id(),
                self.dt.id(),
                mspace.id(),
                fspace.id(),
                self.xfer.id(),
                buf.as_mut_ptr().cast()
            ))?;
            self.pos += amt as u64;
            Ok(out_shape[0])
        }()
        .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err.to_string()))
```
and we can get rid of the trait

src/hl/container.rs
```diff
@@ -346,6 +347,115 @@ impl<'a> Writer<'a> {
...
+        let obj_space = obj.space()?;
+        ensure!(obj_space.shape().len() == 1, "Only rank 1 datasets can be read vie ByteReader.");
+        let xfer = PropertyList::from_id(h5call!(H5Pcreate(*crate::globals::H5P_DATASET_XFER))?)?;
+        if !hdf5_types::USING_H5_ALLOCATOR {
+            crate::hl::plist::set_vlen_manager_libc(xfer.id())?;
+        }
+        Ok(ByteReader { obj, pos: 0, obj_space, dt: mem_dtype, xfer })
+    }
+
+    pub fn set_position(&mut self, pos: u64) {
```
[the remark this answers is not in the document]
Rikorose: Hm I just added this as an equivalent to io::Cursor where it is public. But I guess, using the seek interface is better. I could also remove this method.

src/hl/container.rs
Rikorose: Hm I am not a big fan of that, but as you like. I guess I would rather use the map_err on every io::Result and not create a closure.

src/hl/container.rs
```diff
@@ -346,6 +347,115 @@ impl<'a> Writer<'a> {
...
+impl<T> IOResultExt<T> for Result<T> {
+    fn to_io_err(self) -> std::io::Result<T> {
+        match self {
+            Ok(x) => Ok(x),
+            Err(e) => Err(std::io::Error::new(std::io::ErrorKind::Other, e.to_string())),
+        }
+    }
+}
+
+impl<'a> std::io::Read for ByteReader<'a> {
```
aldanor, line 405: `std::io` -> `io`, here and in other places, since `std::io` is already imported above

src/hl/container.rs
aldanor: 1. Why can't we just implement `From<hdf5::Error> for io::Error` in `error.rs` then?
2. Why `.to_string()`? If I understand correctly, `std::io::Error` will happily box a dyn trait object

src/hl/container.rs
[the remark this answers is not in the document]
aldanor: If you want to make it like `io::Cursor`, then I guess there should also be a public `position()` method. I'm not sure whether copying its interface is useful, but I guess it may be

src/hl/container.rs
```diff
@@ -346,6 +347,115 @@ impl<'a> Writer<'a> {
...
+impl<'a> ByteReader<'a> {
+    pub fn new(obj: &'a Container) -> Result<Self> {
+        ensure!(!obj.is_attr(), "ByteReader cannot be used on attribute datasets");
+
+        let file_dtype = obj.dtype()?;
+        let mem_dtype = Datatype::from_type::<u8>()?;
+        file_dtype.ensure_convertible(&mem_dtype, Conversion::NoOp)?;
+
+        let obj_space = obj.space()?;
+        ensure!(obj_space.shape().len() == 1, "Only rank 1 datasets can be read vie ByteReader.");
```
aldanor, line 368: - Typo: `vie` -> `via`
- Don't need period at the end of error messages

aldanor, line 368: Do we really care if it's rank 1 or rank 2? What's stopping from reading rank N dataset?

tests/test_dataset.rs
```diff
@@ -168,6 +169,83 @@ where
...
+        let n_read = reader.read(&mut chunk).expect("io::Read faild");
+        if pos + chunk_len < arr.len() {
+            // We did not read until end. Thus, the chunk should be fully filled.
+            assert_eq!(chunk_len, n_read);
+        }
+        assert_eq!(&chunk[..n_read], arr.slice(s![pos..pos + n_read]).as_slice().unwrap());
+        pos += chunk_len;
+    }
+
+    // Seek to begining and read again
```
aldanor, line 202: `begining` -> `the beginning`

tests/test_dataset.rs
```diff
@@ -168,6 +169,83 @@ where
...
+    reader.seek(SeekFrom::Start(0)).expect("io::Seek failed");
+    let mut out_bytes = vec![0u8; arr.len()];
+    reader.read(&mut out_bytes.as_mut_slice()).expect("io::Read faild");
```
aldanor, line 205: `faild` -> `failed`

tests/test_dataset.rs
```diff
@@ -168,6 +169,83 @@ where
...
+    let seeked_pos = reader.seek(SeekFrom::End(pos)).expect("io::Seek failed") as usize;
+    assert_eq!(pos, seeked_pos as i64 - arr.len() as i64);
+    let mut out_bytes = vec![0u8; arr.len() - seeked_pos];
+    reader.read(&mut out_bytes.as_mut_slice()).expect("io::Read failed");
+    assert_eq!(out_bytes.as_slice(), arr.slice(s![seeked_pos..]).as_slice().unwrap());
+
+    // Seek before start
+    assert!(reader.seek(SeekFrom::End(-(arr.len() as i64) - 1)).is_err());
+
+    // Test strem position start
```
aldanor, line 240: `strem` -> `stream`

src/hl/container.rs
```diff
@@ -395,6 +505,13 @@ impl Container {
         Writer::new(self)
     }
 
+    /// Creates a ``ByteReader`` which implements ``std::io::{Read, Seek}``.
+    /// ``ByteReader`` only supports 1D ``u8`` datasets. Attributes or datasets
+    /// with other dimensions or datatypes cannot be read with ``ByteReader``.
```
aldanor, line 510:
```suggestion
    /// Creates `ByteReader` which implements [`Read`](std::io::Read)
    /// and [`Seek`](std::io::Seek).
    ///
    /// ``ByteReader`` only supports 1-D `u8` datasets.
```

src/hl/container.rs
```diff
@@ -346,6 +347,115 @@ impl<'a> Writer<'a> {
...
+    fn remaining_len(&self) -> usize {
+        self.dataset_len().saturating_sub(self.pos as usize)
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.pos >= self.dataset_len() as u64
+    }
+}
+
+trait IOResultExt<T> {
```
aldanor, line 393: If we implement `From<Error> for io::Error`, this trait can be removed?

src/hl/container.rs
Rikorose: My intuition was that a reader using ByteReader will have no information about the shape and stride. Therefore, it doesn't really make sense. Or do you mean using a ByteReader on a slice?

src/hl/container.rs
Rikorose: True, I'll add it to src/error.rs

src/hl/container.rs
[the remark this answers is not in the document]
Rikorose: I removed it for now. If we need it, we can add it later.

src/hl/container.rs
mulimoen: `Seek` would be quite odd for a higher dimensioned dataset, and would be really odd if the dataset was resized. We could remove this restriction in the future if we decide on semantics for such iterations

Rikorose:
@aldanor Any update on this PR?

aldanor:
@Rikorose I'm travelling right now, but as soon as I'm back to my computer this weekend / next Monday I'll review this again - hope that's ok.

Rikorose:
Rebased on master

src/hl/container.rs
```diff
@@ -346,6 +347,98 @@ impl<'a> Writer<'a> {
     }
 }
 
+#[derive(Debug)]
+pub struct ByteReader<'a> {
+    obj: &'a Container,
```
mulimoen, line 352: Is it easier to work with the ByteReader if it is owned? Clone is cheap as the object is refcounted

Rikorose: Sure, I guess it makes sense. Then the `Container` needs to be cloned on construction, correct?

mulimoen: Yes, that sounds right

tests/test_dataset.rs
```diff
@@ -171,6 +172,88 @@ where
     Ok(())
 }
 
+fn test_byte_read_seek_impl(ds: &hdf5::Dataset, arr: &ArrayD<u8>, ndim: usize) -> hdf5::Result<()> {
+    let mut rng = SmallRng::seed_from_u64(42);
+    ds.write(arr)?;
+
+    // Read whole
+    let ds = ds.clone();
+    let mut reader = match ds.into_byte_reader() {
```
Rikorose, line 181: This currently fails with
```
error[E0507]: cannot move out of dereference of `_IMPL_H5TYPE_FOR_Enum::_h5::Dataset`
   --> tests/test_dataset.rs:181:28
    |
181 |     let mut reader = match ds.into_byte_reader() {
    |                            ^^^^^^^^^^^^^^^^^^^^^ move occurs because value has type `Container`, which does not implement the `Copy` trait
```
Not quite sure what I am doing wrong.

mulimoen: `Container` is hidden behind the deref trait which is a reference. Either do a double deref before clone (`let ds = (**ds).clone()`) or change signature from `into_byte_reader(self)` to `as_byte_reader(&self)` (which I prefer in this case).

Rikorose: So then I need to do `(**ds).clone()` in `as_byte_reader()`?

mulimoen: It is enough to do a regular clone in `as_byte_reader` as the deref has triggered at the start of the function.

mulimoen (approved):

PR #199, opened 2022-02-03
Add examples

preiter93:
Hi,

I think it would be nice to add a few more examples to expose the most important functions
more prominently. Especially the mpi example could be usefull for some users.

### DIFF
--- a/examples/read_write_ndarray_mpi.rs
+++ b/examples/read_write_ndarray_mpi.rs
@@ -1,0 +1,137 @@
+//! Read and write dynamically sized arrays in parallel.
+//!
+//! This will fail if the systems `hdf5` library is not build
+//! with mpi support.
+//!
+//! ``` ignore
+//! cargo build --example read_write_ndarray_mpi --features mpio
+//! mpirun -np 2 ./target/debug/examples/read_write_ndarray_mpi
+//! ```
+
+static FILE_NAME: &str = "out_read_write_ndarray_mpi.h5";
+static DSET_NAME: &str = "data";
+
+/// Read slice of an ndarray dataset
+///
+/// # Errors
+/// File or dataset do not exist.
+///
+/// # Panics
+/// Panics when `into_dimensionality` fails.
+///
+/// # TODO
+/// Check that slices do not overlap.
+pub fn read_ndarray_mpi<A, D, T>(
+    comm: mpi_sys::MPI_Comm, filename: &str, dsetname: &str, slice: ndarray::SliceInfo<T, D, D>,
+) -> hdf5::Result<ndarray::Array<A, D>>
+where
+    A: hdf5::H5Type,
+    D: ndarray::Dimension,
+    T: AsRef<[ndarray::SliceInfoElem]>,
+{
+    // Open file
+    let file = hdf5::FileBuilder::new().with_fapl(|p| p.mpio(comm, None)).open(filename)?;
+
+    //Read dataset
+    let data = file.dataset(dsetname)?;
+    let y: ndarray::ArrayD<A> = data.read_slice(slice)?;
+
+    // Dyn to static
+    let x = y.into_dimensionality::<D>().unwrap();
+    Ok(x)
+}
+
+/// Write a slice of an ndarray dataset to an
+/// hdf5 file. Specify the full shape of the
+/// dataset and the slice (with ndarrays `s!` macro).
+/// Supplied Array and slices must be smaller than `full_shape`.
+///
+/// Creates new file or append to existing file.
+///
+/// # Errors
+/// File does not exist or file and dataset exist,
+/// but shapes mismatch.
+///
+/// # Panics
+/// Mismatch of number of dimensions between array
+/// and `full_shape`.
+///
+/// # TODO
+/// Check that slices do not overlap.
+pub fn write_ndarray_mpi<A, S, D, T, Sh>(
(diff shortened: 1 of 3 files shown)


examples/read_write_ndarray_mpi.rs
```diff
@@ -0,0 +1,137 @@
...
+///
+/// # TODO
+/// Check that slices do not overlap.
+pub fn read_ndarray_mpi<A, D, T>(
+    comm: mpi_sys::MPI_Comm, filename: &str, dsetname: &str, slice: ndarray::SliceInfo<T, D, D>,
+) -> hdf5::Result<ndarray::Array<A, D>>
+where
+    A: hdf5::H5Type,
+    D: ndarray::Dimension,
+    T: AsRef<[ndarray::SliceInfoElem]>,
```
mulimoen, line 30: Try some variant of `TryInto<Selection>` here for idiomatic code

examples/read_write_ndarray_mpi.rs
```diff
@@ -0,0 +1,137 @@
...
+{
+    // Open file
+    let file = hdf5::FileBuilder::new().with_fapl(|p| p.mpio(comm, None)).open(filename)?;
```
mulimoen, line 33: Move the opening of the file outside of the `read` function

examples/read_write_ndarray_mpi.rs
```diff
@@ -0,0 +1,137 @@
...
+    slice: ndarray::SliceInfo<T, D, D>, full_shape: Sh,
+) -> hdf5::Result<()>
+where
+    A: hdf5::H5Type,
+    S: ndarray::Data<Elem = A>,
+    D: ndarray::Dimension,
+    T: AsRef<[ndarray::SliceInfoElem]>,
+    Sh: Into<hdf5::Extents>,
+{
+    use std::convert::TryFrom;
```
mulimoen, line 72: Move to top of file

examples/read_write_ndarray_mpi.rs
```diff
@@ -0,0 +1,137 @@
...
+    let full_shape = full_shape.into();
+    assert!(
+        array.ndim() == full_shape.ndim(),
+        "Dimension mismatch of array and full_shape, {} vs. {}",
+        array.ndim(),
+        full_shape.ndim()
+    );
+
+    // Create file with mpi access
+    let file = hdf5::FileBuilder::new().with_fapl(|p| p.mpio(comm, None)).append(filename)?;
```
mulimoen, line 83: Move outside of the function

examples/read_write_ndarray_mpi.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    //Write dataset
+    let dset = match file.dataset(dsetname) {
+        // Overwrite
+        Ok(dset) => dset,
+        // Create new dataset
+        std::prelude::v1::Err(..) => {
```
mulimoen, line 90: Drop `std::prelude`, `Err(_)` should match here

mulimoen: Or maybe just unwrap/? if the file creation is moved outside of the function

examples/read_write_ndarray_mpi.rs
```diff
@@ -0,0 +1,137 @@
...
+use std::ptr;
+
+fn main() -> hdf5::Result<()> {
+    // Initialize mpi
+    let mut initialized: c_int = 1;
+    unsafe { mpi_sys::MPI_Initialized(&mut initialized) };
+    if initialized == 0 {
+        unsafe { mpi_sys::MPI_Init(ptr::null_mut(), ptr::null_mut()) };
+    }
+    let world_comm = unsafe { mpi_sys::RSMPI_COMM_WORLD };
```
mulimoen, line 108: Unnecssary `unsafe`?

examples/read_write_ndarray_mpi.rs
```diff
@@ -0,0 +1,137 @@
...
+    let mut rank: c_int = 1;
+    unsafe { mpi_sys::MPI_Comm_rank(world_comm, &mut rank) };
```
mulimoen, line 110: You should assert `world-size == 2`

mulimoen (commented):
Great to see another example! It would be great if these two examples could be merged into one; where the MPI isage is shown side-by-side the "normal" usage. I think `mpi` is fairly abstracted away and the differences could be made fairly minor

preiter93:
Thank you for the review!

I have removed the previous examples and added two new ones, one for serial read/write
and the other for mpi read/write. I think it is better to separate serial and parallel.

I'm not quite sure when to use file.close(). For the mpi example it will fail, so I'll let the close step out.

And feel free to edit the examples, I have only superficial knowledge with this crate.

examples/read_write_ndarray_mpi.rs
preiter93: seems necessary (use of mutable static)

mulimoen:
@preiter93 I had a small go over here: https://github.com/mulimoen/hdf5-rust/pull/15, do these modifications make sense in how you want to examplify the crate?

preiter93:
@mulimoen Yes looks very nice now. Just made two tiny changes:

- #[cfg(mpio)] -> #[cfg(feature = "mpio")]
- Remove lifetime of read

PR #203, opened 2022-02-18
Add hdf5 helper attribute with rename functionality

### LINKED ISSUE #201: Specify field names in compound types
I have an HDF5 dataset consisting of a 1D array of compound types, produced by astropy on a table with a `MaskedColumn`. The field name for the mask is "foo.mask", which does not seem to play nice with the derive macro. It would be useful to support something like:

```rust
#[derive(H5Type, Debug)]
#[repr(C)]
struct MyRow {
    foo: f64,
    #[hdf5(name = "foo.mask")]
    foo_mask: bool,
}

// ...

let array = dataset.read_1d::<MyRow>()?;
```

Apologies if this kind of thing is already supported and I didn't manage to find the right place in the documentation.

ajtribick:
Initial implementation of the helper attribute to allow field renaming. Resolves #201

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Support for HDF5 version 1.13.0.
+- Support field renaming via `#[hdf5(rename = "new_name")]` helper attribute.
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


mulimoen:
This looks really great. It seems very straight forward to apply the same method to enums too if you change the call signature of `extract_field_name` to take `&[Attribute]` and return `Option<String>`. Feel free to add to the changelog too, this is a very useful feature.

mulimoen (approved):
Looks great, any objections against merging this @aldanor?

ajtribick:
Any further changes needed here?

mulimoen:
Looks great, thanks for the PR!

PR #204, opened 2022-02-19
Add missing proxy features to hdf5 crate

ajtribick:
Proxy features from hdf5-sys crate, since cargo doesn't allow the following:

```toml
[dependencies]
hdf5 = { version = "0.8.1", features = ["hdf5-sys/static"] }
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with: {toolchain: nightly, profile: minimal, override: true}
       - name: Document workspace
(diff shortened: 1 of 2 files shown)


mulimoen:
I think this should wait till we bump MSRV to `1.60` and use weak dependencies. This way it will be less confusing if one specifies `zlib`, if done as of now this will compile and use the included version of `hdf5-c`.

We have to specify what a static build should mean. In some instances the `hdf5-c` library is packaged as a static library which could be used directly. Ought we rename this flag to `build-from-source` to make this less ambigous?

The `hl`, `threadsafe`, `deprecated` options are only necessary for users that use `hdf5-sys` indirectly (such as `georust/netcdf`), and they need not be proxied.

Finally: This has a workaround which is including `hdf5-sys` yourself in your `Cargo.toml` and specifying exactly how you want the library built.

PR #207, opened 2022-03-23
WIP: Add multiple hdf5 versions for static building

mulimoen:


mulimoen:
@aldanor This PR also fixes CI

mulimoen:
@aldanor It would be great if you could have a look at this!

mulimoen:
I replaced hdf5 1.13 (which was a beta lineage) with 1.14.0. This requires #227

aldanor:
Do we want to go ahead with this? If so, 1.13->1.14 and will need a rebase?

mulimoen:
Would be easier to bump the bundled `hdf5-c` to 1.14 instead of maintaining three versions

aldanor:
Probably, yea... Now that 1.14 is stable, is out on most platforms and package managers and will be a default version of choice for a long while anyway, I guess.

PR #209, opened 2022-04-06
Call silence_errors for every thread

mulimoen:
Fixes error messages not being silenced in a multi-threaded setting as mentioned in https://github.com/aldanor/hdf5-rust/issues/155#issuecomment-1090189833


aldanor:
- Looks good!
- Mention in the changelog?

(on a side note, I think we should switch to using `once_cell`... I guess this will be a part of a big overhaul once I get to it)

aldanor (approved):

mulimoen:
Added an entry to the changelog

aldanor:
I guess clippy needs to be fixed first in one of the PRs so we can rebase the rest on it and the builds are green?

PR #210, opened 2022-04-08
Implement H5Type for complex

### LINKED ISSUE #208: Store num_complex datatypes
Hi,

I am wondering what the best way is to store a `Complex<f32>` in a hdf5 dataset.

Since the type is defined in the num_complex crate, I cannot easily derive H5Type. Do I need to define a wrapper type? Or, can this be added to hdf5-rust maybe as a Cargo feature?

What other options do I have?

mulimoen:
Closes #208

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Add a `ByteReader` which implements `std::io::{Read, Seek}` for 1D `u8`
   datasets. Usage via `Dataset::as_byte_reader()`.
 - Add `chunk_visit` to visit all chunks in a dataset.
+- Implement `H5Type` for `num_complex::Complex`.
 
 ### Changed
 
--- a/hdf5-types/Cargo.toml
+++ b/hdf5-types/Cargo.toml
@@ -15,12 +15,14 @@ edition.workspace = true
 
 [features]
 h5-alloc = []
+complex = ["num-complex"]
 
 [dependencies]
 ascii = "1.1"
 cfg-if = { workspace = true }
 hdf5-sys = { workspace = true }
 libc = { workspace = true }
+num-complex = { version = "0.4", optional = true, default-features = false }
 
 [dev-dependencies]
 quickcheck = { version = "1.0", default-features = false }
--- a/hdf5-types/src/complex.rs
+++ b/hdf5-types/src/complex.rs
@@ -6,6 +6,8 @@ use num_complex::Complex;
 
 unsafe impl<T: H5Type> H5Type for Complex<T> {
     fn type_descriptor() -> TypeDescriptor {
+        // Complex<T> should be FFI-equivalent to [T; 2]
+        // https://docs.rs/num-complex/0.4.3/num_complex/struct.Complex.html#representation-and-foreign-function-interface-compatibility
         TypeDescriptor::Compound(CompoundType {
             fields: vec![
                 // Compatible with h5py definition of complex
--- a/hdf5-types/src/complex.rs
+++ b/hdf5-types/src/complex.rs
@@ -1,0 +1,30 @@
+use super::{CompoundField, CompoundType, H5Type, TypeDescriptor};
+
+use std::mem::size_of;
+
+use num_complex::Complex;
+
+unsafe impl<T: H5Type> H5Type for Complex<T> {
+    fn type_descriptor() -> TypeDescriptor {
+        TypeDescriptor::Compound(CompoundType {
+            fields: vec![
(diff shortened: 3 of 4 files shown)


hdf5-types/Cargo.toml
```diff
@@ -14,12 +14,14 @@ categories = ["encoding"]
...
 [features]
 h5-alloc = []
+complex = ["num-complex"]
 
 [dependencies]
 ascii = "1.0"
 libc = "0.2"
 hdf5-sys = { version = "0.8.1", path = "../hdf5-sys" }  # !V
 cfg-if = "1.0.0"
+num-complex = { version = "0.4.0", optional = true, default-features = false }
```
aldanor, line 24: @mulimoen Maybe set to `"0.4"`? E.g. because 0.4.2 is already out.

hdf5-types/src/complex.rs
```diff
@@ -0,0 +1,30 @@
+use super::{CompoundField, CompoundType, H5Type, TypeDescriptor};
+
+use std::mem::size_of;
+
+use num_complex::Complex;
+
+unsafe impl<T: H5Type> H5Type for Complex<T> {
+    fn type_descriptor() -> TypeDescriptor {
+        TypeDescriptor::Compound(CompoundType {
```
aldanor, line 9: This is inherently unsafe, how do we know that offsets are right? (e.g. should we check actual offsets here?)

mulimoen: I am not sure how we would check the offsets

hdf5-types/Cargo.toml
mulimoen: Set to 0.4

hdf5-types/src/complex.rs
mulimoen: That would require us to make something to point at. We could make an instance of `MaybeUninit`, but field access is only a future feature

aldanor: Hm... I guess we could do something like this, does that make sense? https://rust.godbolt.org/z/TWbavj98G

mulimoen: We could do
```rust
unsafe impl<T: H5Type> H5Type for Complex<T> {
    fn type_descriptor() -> TypeDescriptor {
        use std::mem::MaybeUninit;
        // MaybeUninit has the same size and is transparent, but we can
        // do some pointer arithmetics on a real instance
        let dummystruct: Complex<MaybeUninit<T>> =
            Complex::new(MaybeUninit::uninit(), MaybeUninit::uninit());

        assert_eq!(size_of_val(&dummystruct), size_of::<Complex<T>>());

        let base = addr_of!(dummystruct).cast::<u8>();
        let ptr_r = addr_of!(dummystruct.re).cast::<u8>();
        let ptr_i = addr_of!(dummystruct.im).cast::<u8>();

        TypeDescriptor::Compound(CompoundType {
            fields: vec![
                // Compatible with h5py definition of complex
                CompoundField::typed::<T>("r", unsafe { ptr_r.offset_from(base) } as usize, 0),
                CompoundField::typed::<T>("i", unsafe { ptr_i.offset_from(base) } as usize, 1),
            ],
            size: size_of::<Complex<T>>(),
        })
    }
}
```

mulimoen: Might be enough to link to the documentation on representation? https://docs.rs/num-complex/0.4.2/num_complex/struct.Complex.html#representation-and-foreign-function-interface-compatibility

aldanor: Maybe add this link in a comment in code? E.g. "`Complex<T>` should be FFI-equivalent to `[T; 2]`: <link>".

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Support field renaming via `#[hdf5(rename = "new_name")]` helper attribute.
 - Add a `ByteReader` which implements `std::io::{Read, Seek}` for 1D `u8`
   datasets. Usage via `Dataset::as_byte_reader()`.
+- `H5Type` for `Complex` struct from `num_complex`
```
aldanor, line 12:
```suggestion
- Implement `H5Type` for `num_complex::Complex`.
```

aldanor:
Minor nits aside, needs to be rebased on master (because of the changelog).

hdf5-types/src/complex.rs
mulimoen: Thanks, added this as a comment under `type_descriptor()`

aldanor (approved):

PR #213, opened 2022-04-18
Add check for chunking of dataset for filters

### LINKED ISSUE #212: `write_scalar` causes unknown library error
I add a simple test to my fork. It passes on my local computer but fails on many platforms, see CI results here: https://github.com/kaizhang/hdf5-rust/actions/runs/2168628529

The test code is:

```rust
#[cfg(test)]
mod tests {
    use super::*;
    use crate::test::with_tmp_file;

    #[test]
    fn test_read_write_scalar() {
        with_tmp_file::<Result<_>, _>(|file| {
            let val: f64 = 0.2;
            let dataset = file.new_dataset::<f64>().deflate(3).create("foo")?;
            dataset.write_scalar(&val)?;
            let val_back = dataset.read_scalar()?;
            assert_eq!(val, val_back);
            Ok(())
        })
        .unwrap()
    }
}
```

mulimoen:
The behaviour in `h5py` is to raise an error when trying to apply chunking/compression filters to dataset. This PR adds the same behaviour.

Fixes #212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 
 - The `H5Type` derive macro now uses `proc-macro-error` to emit error messages.
 - MSRV is now `1.64.0` and Rust edition has now been bumped to 2021.
-- Types in ChunkInfo has been changed to match HDF5
+- Types in ChunkInfo has been changed to match HDF5.
 
 ### Fixed
 
@@ -25,6 +25,7 @@
 - Fixed a bug where errors were only silenced on the main thread.
 - Fixed a memory leak when opening datasets.
 - Avoid creating unaligned references in `H5Type` derive macro.
+- Applying filters without chunking will now produce an explicit error.
 
 ## 0.8.1
 
--- a/hdf5/src/hl/filters.rs
+++ b/hdf5/src/hl/filters.rs
@@ -517,11 +517,15 @@ impl Filter {
             Ok(filters)
         })
     }
+
+    pub(crate) fn requires_chunking(&self) -> bool {
+        COMP_FILTER_IDS.contains(&self.id())
+    }
 }
 
-pub(crate) fn validate_filters(filters: &[Filter], type_class: H5T_class_t) -> Result<()> {
-    const COMP_FILTER_IDS: &[H5Z_filter_t] = &[H5Z_FILTER_DEFLATE, H5Z_FILTER_SZIP, 32000, 32001];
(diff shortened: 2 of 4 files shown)


hdf5/src/hl/filters.rs
```diff
@@ -517,11 +517,15 @@ impl Filter {
             Ok(filters)
         })
     }
+
+    pub(crate) fn requires_chunking(&self) -> bool {
+        COMP_FILTER_IDS.contains(&self.id())
```
aldanor, line 522: Just wondering, that's what h5py does as well? I.e. just check a list of hardcoded well-known filter ids (because obviously there's other third party ones...)

I guess there's not much else you can do because there's no supplied filter metadata

mulimoen: I am not sure if a user-supplied filter needs to have filtering? The behaviour if the filter is not known is to pass the error through, worst case the user gets a confusing error

aldanor: I think this is pretty much the whole list of what people actually use: https://github.com/silx-kit/hdf5plugin/tree/main/src

mulimoen: The list is different from ours but is similar https://github.com/h5py/h5py/blob/49d7b8cde724a69a50d6cd4b218df7a6ba141633/h5py/_hl/dataset.py#L89

aldanor: Should we do the same (+blosc) then?

Although I'm not entirely sure all the non-compression filters they have listed _actually_ require chunking 🤔

mulimoen: From https://support.hdfgroup.org/HDF5/doc/H5.user/Filters.html it seems all filters build on chunked data, so the only valid `required_chunking` is the one that returns `true`

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - Fixed a bug where errors were only silenced on the main thread.
 - Fixed a memory leak when opening datasets.
 - Avoid creating unaligned references in `H5Type` derive macro.
+- Fixed compression filter not giving an error on non-chunked datasets.
```
aldanor, line 25: Now that we've changed the check, maybe reword it as "Applying filters without chunking will now produce an explicit error."?

aldanor:
Could you please rebase on master now that there's new stuff merged in?

CHANGELOG.md
mulimoen: Fixed, thanks for keeping this up to date

aldanor (approved):

PR #214, opened 2022-05-03
Chunks iterator

mulimoen:
This exposes the new chunk iterator in hdf5 1.14

### DIFF
--- a/hdf5/src/hl/chunks.rs
+++ b/hdf5/src/hl/chunks.rs
@@ -1,89 +1,165 @@
 use crate::internal_prelude::*;
-use hdf5_sys::h5d::H5Dchunk_iter;
 
-/// Borrowed version of [ChunkInfo](crate::dataset::ChunkInfo)
-#[derive(Debug)]
-pub struct ChunkInfoBorrowed<'a> {
-    pub offset: &'a [u64],
+#[cfg(feature = "1.10.5")]
+use hdf5_sys::h5d::{H5Dget_chunk_info, H5Dget_num_chunks};
+
+#[cfg(feature = "1.10.5")]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ChunkInfo {
+    /// Array with a size equal to the dataset’s rank whose elements contain 0-based
+    /// logical positions of the chunk’s first element in each dimension.
+    pub offset: Vec<u64>,
+    /// Filter mask that indicates which filters were used with the chunk when written.
+    /// A zero value indicates that all enabled filters are applied on the chunk.
+    /// A filter is skipped if the bit corresponding to the filter’s position in
+    /// the pipeline (0 ≤ position < 32) is turned on.
     pub filter_mask: u32,
+    /// Chunk address in the file.
     pub addr: u64,
+    /// Chunk size in bytes.
     pub size: u64,
 }
 
-struct RustCallback<F> {
-    ndims: usize,
-    callback: F,
-}
+#[cfg(feature = "1.10.5")]
+impl ChunkInfo {
+    pub(crate) fn new(ndim: usize) -> Self {
+        let offset = vec![0; ndim];
+        Self { offset, filter_mask: 0, addr: 0, size: 0 }
+    }
 
-extern "C" fn chunks_callback<F>(
-    offset: *const hsize_t, filter_mask: u32, addr: haddr_t, nbytes: u32, op_data: *mut c_void,
-) -> herr_t
-where
-    F: FnMut(ChunkInfoBorrowed) -> i32,
-{
-    unsafe {
-        std::panic::catch_unwind(|| {
-            let data: *mut RustCallback<F> = op_data.cast::<RustCallback<F>>();
-            let ndims = (*data).ndims;
-            let callback = &mut (*data).callback;
-
-            let offset = std::slice::from_raw_parts(offset, ndims);
-
-            let info =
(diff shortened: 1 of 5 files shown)


hdf5/src/hl/chunks.rs
```diff
@@ -0,0 +1,181 @@
...
+use hdf5_sys::h5d::{H5Dget_chunk_info, H5Dget_num_chunks};
+
+#[cfg(feature = "1.10.5")]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ChunkInfo {
+    /// Array with a size equal to the dataset’s rank whose elements contain 0-based
+    /// logical positions of the chunk’s first element in each dimension.
+    pub offset: Vec<u64>,
+    /// Filter mask that indicates which filters were used with the chunk when written.
+    /// A zero value indicates that all enabled filters are applied on the chunk.
```
aldanor, line 13:
```suggestion
    /// Filter mask that indicates which filters were used with the chunk when written.
    ///
    /// A zero value indicates that all enabled filters are applied on the chunk.
```

hdf5/src/hl/chunks.rs
```diff
@@ -0,0 +1,181 @@
...
+        return None;
+    }
+    h5lock!(ds.space().map_or(None, |s| {
+        let mut n: hsize_t = 0;
+        h5check(H5Dget_num_chunks(ds.id(), s.id(), &mut n)).map(|_| n as _).ok()
+    }))
+}
+
+#[cfg(feature = "1.14.0")]
+mod one_thirteen {
```
aldanor, line 69: Something like this would be probably a bit easier to read... (or `v1_13_0`)
```suggestion
mod v1_14_0 {
```

hdf5/src/hl/chunks.rs
```diff
@@ -0,0 +1,181 @@
...
+    use super::*;
+    use hdf5_sys::h5d::H5Dchunk_iter;
+
+    /// Borrowed version of [ChunkInfo](crate::dataset::ChunkInfo)
+    #[derive(Clone, Debug, PartialEq, Eq)]
+    pub struct ChunkInfoBorrowed<'a> {
```
aldanor, line 75:
```suggestion
    pub struct ChunkInfoRef<'a> {
```

(no hard preference but this would be slightly more conventional naming)

hdf5/src/hl/chunks.rs
```diff
@@ -0,0 +1,181 @@
...
+                addr: val.addr,
+                size: val.size,
+            }
+        }
+    }
+
+    struct RustCallback<F> {
+        ndims: usize,
+        callback: F,
+    }
```
aldanor, line 103:
```suggestion
    #[repr(C)]
    struct RustCallback<F> {
        pub ndims: hsize_t,
        pub callback: F,
    }
```

hdf5/src/hl/chunks.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub offset: Vec<u64>,
+    /// Filter mask that indicates which filters were used with the chunk when written.
+    /// A zero value indicates that all enabled filters are applied on the chunk.
+    /// A filter is skipped if the bit corresponding to the filter’s position in
+    /// the pipeline (0 ≤ position < 32) is turned on.
+    pub filter_mask: u32,
+    /// Chunk address in the file.
+    pub addr: u64,
+    /// Chunk size in bytes.
+    pub size: u64,
```
aldanor, line 20: `addr` is clearly `haddr_t` which is _usually_ `u64`, however if I'm reading this correctly, `size` is `hsize_t`, so it should probably be mapped to `size: usize`?

hdf5/src/hl/chunks.rs
```diff
@@ -0,0 +1,181 @@
+use crate::internal_prelude::*;
```
aldanor, line 1: Wondering, is there a good reason for moving this into a `chunks` module? (no hard preference though...)

mulimoen: `dataset.rs` is 1000 lines already so keeping it short in a separate file. I'd be happy to move it back, makes no difference to me

hdf5/src/hl/chunks.rs
mulimoen: You are quite right, have changed this to be consistent with the c library

hdf5/src/hl/chunks.rs
aldanor: Agreed, let's leave it be for now. (ideally, things should be split up in many, many more modules...)

hdf5/src/hl/chunks.rs
```diff
@@ -0,0 +1,182 @@
...
+
+#[cfg(feature = "1.10.5")]
+use hdf5_sys::h5d::{H5Dget_chunk_info, H5Dget_num_chunks};
+
+#[cfg(feature = "1.10.5")]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ChunkInfo {
```
aldanor, line 8: One last thing, maybe a one-liner docstring for the struct itself?

aldanor:
👍 Will merge it after it's green if there's nothing else pending.

aldanor:
We almost forgot the most important part, the changelog entry 🙂

mulimoen:
Added to changelog and noticed a place where we used 1.13 which I've changed to 1.14

aldanor (approved):

PR #216, opened 2022-05-08
Use weak features in Cargo.toml

mulimoen:
Expands on the idea of @ajtribick in #204

### DIFF
--- a/hdf5-sys/Cargo.toml
+++ b/hdf5-sys/Cargo.toml
@@ -23,12 +23,12 @@ hdf5-src = { workspace = true, optional = true }
 # Please see README for further explanation of these feature flags
 [features]
 default = []
-mpio = ["mpi-sys"]
-hl = ["hdf5-src/hl"]
-threadsafe = ["hdf5-src/threadsafe"]
-zlib = ["libz-sys", "hdf5-src/zlib"]
-static = ["hdf5-src"]
-deprecated = ["hdf5-src/deprecated"]
+mpio = ["dep:mpi-sys"]
+hl = ["hdf5-src?/hl"]
+threadsafe = ["hdf5-src?/threadsafe"]
+zlib = ["libz-sys", "hdf5-src?/zlib"]
+static = ["dep:hdf5-src"]
+deprecated = ["hdf5-src?/deprecated"]
 
 [build-dependencies]
 libloading = "0.8"
--- a/hdf5-sys/build.rs
+++ b/hdf5-sys/build.rs
@@ -684,20 +684,18 @@ impl Config {
     }
 
     fn check_against_features_required(&self) {
-        if feature_enabled("DEPRECATED") {
-            assert!(!self.header.have_no_deprecated, "Required deprecated symbols are not present")
-        }
-        if feature_enabled("THREADSAFE") {
-            assert!(
-                self.header.have_threadsafe,
-                "Required threadsafe but library was not build using the threadsafe option"
-            );
-        }
-        if feature_enabled("ZLIB") {
-            assert!(
-                self.header.have_zlib,
-                "Required zlib filter but library does not have builtin support for this options"
-            );
+        let h = &self.header;
+        for (flag, feature, native) in [
+            (h.have_no_deprecated, "deprecated", "HDF5_ENABLE_DEPRECATED_SYMBOLS"),
+            (h.have_threadsafe, "threadsafe", "HDF5_ENABLE_THREADSAFE"),
+            (h.have_zlib, "zlib", "HDF5_ENABLE_Z_LIB_SUPPORT"),
+        ] {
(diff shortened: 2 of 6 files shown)


aldanor:
Rebase on master?

aldanor:
MSRV fail?

aldanor:
> We will have to bump msrv to use weak features. 1.60 was released 7th April 2022

That's problem ok, simplifies maintenance going forward (whoever needs to use older rustc can always use the current crate version).

Probably makes sense to do a few more things:
- update the README (it mentions MSRV)
- list it in Cargo.toml of all crates (https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field)
- bump the edition of all crates to 2021 (and mention it along with MSRV bump in the changelog)
- perhaps also change if any deps are outdated, idk? (parking_lot, mpi_sys, scopeguard, regex, etc... but maybe it could be a separate pr; lazy_static might even be thrown out in favor of once_cell...)

mulimoen:
I updated the dependencies listed by `cargo outdated`

mulimoen:
We could start using [workspace inheritance](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) as this has been stabilized

mulimoen:
Rebased and cleared for review

hdf5-sys/build.rs
```diff
@@ -634,6 +641,24 @@ impl Config {
...
+                self.header.have_threadsafe,
+                "Required threadsafe but library was not build using the threadsafe option"
+            );
+        }
+        if feature_enabled("ZLIB") {
+            assert!(
+                self.header.have_zlib,
+                "Required zlib filter but library does not have builtin support for this options"
+            );
+        }
```
aldanor, line 660: Maybe something like this for consistency of error messages?
```rust
let h = &self.header;
for (flag, feature, native) in [
    (h.have_no_deprecated, "no-deprecated", "H5_NO_DEPRECATED_SYMBOLS"),
    (h.have_threadsafe, "threadsafe", "H5_HAVE_THREADSAFE"),
    (h.have_zlib, "zlib", "H5_HAVE_FILTER_DEFLATE"),
] {
    if feature_enabled(feature.to_ascii_uppercase().replace('-', '_')) {
        assert!(flag, "Enabled feature {feature:?} but the HDF5 library wasn't built with {native}");
    }
}

File: hdf5-sys/Cargo.toml [hunk 2]
```diff
@@ -23,12 +23,12 @@ hdf5-src = { workspace = true, optional = true }
...
-threadsafe = ["hdf5-src/threadsafe"]
-zlib = ["libz-sys", "hdf5-src/zlib"]
-static = ["hdf5-src"]
-deprecated = ["hdf5-src/deprecated"]
+mpio = ["dep:mpi-sys"]
+hl = ["hdf5-src?/hl"]
+threadsafe = ["hdf5-src?/threadsafe"]
+zlib = ["libz-sys", "hdf5-src?/zlib"]
+static = ["dep:hdf5-src"]
+deprecated = ["hdf5-src?/deprecated"]
```
aldanor, line 31: It's the other way around, _**no**_ deprecated symbols.

So it should probably be called `no-deprecated`?

aldanor:
Probably worth adding an entry to the changelog.

(and another entry to Changed if we rename 'deprecated'->'no-deprecated')

hdf5-sys/build.rs
mulimoen: I used your suggestions but instead use the compile time flags which are used for building HDF5 with the feature enabled

hdf5-sys/Cargo.toml
aldanor: So let's rename `deprecated` -> `no-deprecated` while we're changing things here?

mulimoen: In `hdf5-src` we ask for the inclusion of deprecated symbols, I would like to keep this up the chain since asking for no deprecated symbols seems rather counter-intuitive.

hdf5-sys/build.rs
aldanor: > I used your suggestions but instead use the compile time flags which are used for building HDF5 with the feature enabled

That's even better, thanks! (I think that's what I meant really)

hdf5-sys/Cargo.toml
aldanor: Ok, I got a bit confused myself :) But then... shouldn't we be checking `!header.have_no_deprecated` as opposed to `header.have_no_deprecated`?

mulimoen: Yeah, you are right! Bit confusing with the double negation but should be correct now

aldanor:
Ok, I think it should be good to go - I'll merge it when it's green 👍

aldanor (approved):

PR #219, opened 2022-09-05
Use H5Pisa_class to circumvent bug

### LINKED ISSUE #218: Memory Leak in Dataset Creation
The following code will leak about 120 MB. Changing `GROUPS` to `400000` leaks 1.2 GB. `heaptrack` suggests the problem lies in creating the plists for the dataset. I've tried calling the underlying hdf5 close functions on the file, group, dataset and plist without solving it, as well as trying to call the underlying hdf5 garbage collection function.

```
use hdf5;

const PER_GROUP: u64 = 10;
const GROUPS: u64 = 40000;


fn main() {
    let fake_data: [u16; 512] = [255; 512];
    //Create file
    let hdffile = hdf5::File::create("AFile.h5").unwrap();
    
        
    for i in 0..GROUPS
    {
        let groupname = format!("{}", i);
        //Create group
        let hdfgroup = hdffile.create_group(&groupname).unwrap();
        

        for i in 0..PER_GROUP {
            //Write PER_GROUP 8kb arrays as individual datasets
            let ds_name = format!("{}", i);

            hdfgroup
                .new_dataset_builder()
                .with_data(&fake_data)
                .create(&*ds_name)
                .unwrap();

            /*heaptrack will show that the call to create will cause the majority
            of memory leaks, when it creates the plist for access properties*/
        }
    }
    //Close the HDF file
    hdffile.close().unwrap();
    println!("Finished!");
    loop {
        //Hang about until quit.
        //Memory useage will remain high, would expect dropping the file to stop that
        std::hint::spin_loop();
    }
}
```

mulimoen:
Fixes #218. On the the bug in the issue memory usage decreases from 100MB++ to less than 50kB

@stormmathisen and @smutch Would be great if you tested this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
   with feature `have-parallel` instead of `have-direct`.
 - Fixed a missing symbol when building `hdf5-src` with `libz-sys`.
 - Fixed a bug where errors were only silenced on the main thread.
+- Fixed a memory leak when opening datasets.
 
 ## 0.8.1
 
--- a/src/hl/plist.rs
+++ b/src/hl/plist.rs
@@ -5,8 +5,8 @@ use std::ptr::{self, addr_of_mut};
 use std::str::FromStr;
 
 use hdf5_sys::h5p::{
-    H5Pcopy, H5Pequal, H5Pexist, H5Pget_class, H5Pget_class_name, H5Pget_nprops, H5Piterate,
-    H5Pset_vlen_mem_manager,
+    H5Pcopy, H5Pequal, H5Pexist, H5Pget_class, H5Pget_class_name, H5Pget_nprops, H5Pisa_class,
+    H5Piterate, H5Pset_vlen_mem_manager,
 };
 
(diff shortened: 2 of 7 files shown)


smutch:
Hi @mulimoen. Thanks for looking in to this.

I can confirm a noticeable improvement for my small test run (3,819 groups with ~11,500 datasets).

Before:
Screen Shot 2022-09-06 at 10 57 00

After:
Screen Shot 2022-09-06 at 10 57 24

There is a large reduction in persistent 96 byte allocations.

aldanor:
- Probably makes sense to fix a few clippy complaints as well here?
- Add this to changelog?
- Why do you say it "partially" fixes #218? Wasn't this the root of the problem?

mulimoen:
There are enough breaking clippy complaints that I've opened #220 for those.

Added an entry to the changelog and changed the description of this PR

aldanor:
Rebase on master?

aldanor (approved):

aldanor:
👍

PR #226, opened 2023-01-15
Allow chunking on zero-sized extents

### LINKED ISSUE #223: Chunking required when filters are present
Hey guys - I'm using an external HDF5 filter to compress and chunk data when writing to a HDF5 file, specifically [VBZ_compression](https://github.com/nanoporetech/vbz_compression). 

In order to do this I set the `HDF5_PLUGIN_PATH` as so 
    std::env::set_var("HDF5_PLUGIN_PATH", "./vbz_plugin".resolve().as_os_str());

and when I write the compressed/chunked data like so 

```rust
#[cfg(feature = "blosc")]
blosc_set_nthreads(2); // set number of blosc threads
let builder = group.new_dataset_builder();
builder
    .add_filter(32020, &[0, 2, 1, 1])
    .with_data(&data)
    .create("Signal")?;
```
Data here is a `Vec<i16>` and group is a ` hdf5::Group`.

It works about 99.9% of the time. However it very rarely throws an `Err` value - Chunking required when filters are present. Would this behavior potentially occur if the signal `Vec` were smaller than the chunk size, or if it were empty?

The error specifically comes from here :
https://github.com/aldanor/hdf5-rust/blob/fb96296ee3e44e8584f888f89ec2d0693e26c6bd/src/hl/dataset.rs#L447

Can anyone help me work out why this happens? Should I just catch this error and try again?

Many thanks,

Rory

mulimoen:
Fixes #223

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - Fixed a memory leak when opening datasets.
 - Avoid creating unaligned references in `H5Type` derive macro.
 - Applying filters without chunking will now produce an explicit error.
+- Fixed a bug where chunking could not be enabled for zero-sized extents.
 
 ## 0.8.1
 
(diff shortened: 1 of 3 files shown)


aldanor:
Ok, this looks simple/reasonable. Rebase + changelog update?

(btw: #223 this links to has been closed by the previously merged PR?)

aldanor (approved):

aldanor:
👍

PR #227, opened 2023-01-16
Support hdf5 1.14.0

mulimoen:
The work done to support 1.13.0 means 1.14.0 is trivial to support

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,8 @@ jobs:
         include:
           - {version: hdf5@1.8}
           - {version: hdf5@1.10}
-          # - {version: hdf5-mpi, mpi: true}  # TODO: re-enable once 1.14 support is merged in
+          - {version: hdf5@1.14}
+          - {version: hdf5-mpi, mpi: true}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


jan-kaspar:
Hi all, what is the status of this PR? I'd be very grateful for the support of 1.14...

BTW: could one ignore the "patch" part here: https://github.com/aldanor/hdf5-rust/blob/master/hdf5-sys/build.rs#L32 ? For example on ArchLinux there is just a number after the dash, https://archlinux.org/packages/community/x86_64/hdf5/, thus failing the regex.

Many thanks in advance!

aldanor:
@mulimoen I'll take a look at this (and other PRs) on the weekend, is h5d the only change here?

mulimoen:
@aldanor Yes, everything else is covered by the support for 1.13

We should probably remove support for 1.13 at some point considering odd minor versions are "test" versions

mulimoen:
Patched versions in e.g. Arch linux currently has the version number "1.14.1-2" which should now be supported

.github/workflows/ci.yml
```diff
@@ -193,7 +197,7 @@ jobs:
       fail-fast: false
       matrix:
         rust: [stable]
-        version: ["1.8", "1.10", "1.12", "1.13"]
+        version: ["1.8", "1.10", "1.12", "1.14"]
```
aldanor, line 200: Yea, probably the right call to just throw away 1.13 from the CI and replace it with 1.14 where possible.

aldanor:
Looks good, thanks! Unless there's anything else to be done here I'll merge this so we can proceed forward.

aldanor:
Just double checking - would this close https://github.com/aldanor/hdf5-rust/issues/230?

mulimoen:
Feel free to merge. I'll prepare something to rid us of 1.13 later

aldanor (approved):

PR #232, opened 2023-03-07
Enable `T` in `VarLenArray<T>` to also contain a `VarLenArray`

### LINKED ISSUE #222: `T` in `VarLenArray<T>` cannot contain another `VarLenArray`
Hello,
I have a file where one dataset is stored as a `VarLenArray` of a custom compound which contains at least one `VarLenArray`. The following code does not work with the current version as the trait `Copy` is not implemented for `VarLenArray`

```rust
#[derive(H5Type, Copy, Debug)] // register with HDF5
#[repr(C)]
pub struct Cluster {
    id: i32,
    name: VarLenAscii,
    colour: i32,
    neighbours: VarLenArray<i32>,
}

fn read_clusters(dataset: Dataset) -> Result<()> {
    let clusters: VarLenArray<Cluster> = dataset.read_scalar()?;

    println!("{:?}", clusters);

    // ... process clusters ...

    Ok(())
}
```

Results in the error:

```
error[E0204]: the trait `Copy` may not be implemented for this type
  --> src\main.rs:60:18
   |
60 | #[derive(H5Type, Copy, Debug)] // register with HDF5
   |                  ^^^^
...
63 |     name: VarLenAscii,
   |     ----------------- this field does not implement `Copy`
...
67 |     neighbours: VarLenArray<i32>,
   |     -------------------------- this field does not implement `Copy`
```

If I change `VarLenArray` to requiring `Clone` instead of `Copy` then I can get this to compile, run and read in the correct data, but I am unsure what effects that has on the code. What would be a good solution to this problem?

AlanRace:
This fixes #222.

Additionally, I was implementing a custom filter and required access to the h5err macro, hence the changes to the macro exporting.

### DIFF
--- a/hdf5-types/src/array.rs
+++ b/hdf5-types/src/array.rs
@@ -6,13 +6,13 @@ use std::ptr;
 use std::slice;
 
 #[repr(C)]
-pub struct VarLenArray<T: Copy> {
+pub struct VarLenArray<T: Clone> {
     len: usize,
     ptr: *const T,
     tag: PhantomData<T>,
 }
 
-impl<T: Copy> VarLenArray<T> {
+impl<T: Clone> VarLenArray<T> {
     pub unsafe fn from_parts(p: *const T, len: usize) -> Self {
         let (len, ptr) = if !p.is_null() && len != 0 {
             let dst = crate::malloc(len * mem::size_of::<T>());
@@ -50,7 +50,7 @@ impl<T: Copy> VarLenArray<T> {
     }
 }
 
-impl<T: Copy> Drop for VarLenArray<T> {
+impl<T: Clone> Drop for VarLenArray<T> {
     fn drop(&mut self) {
         if !self.ptr.is_null() {
             unsafe {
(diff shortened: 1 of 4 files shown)


aldanor:
This PR contains tons of unrelated changes and will need rebasing and cleaning, however that's not even the main issue here.

You _can't_ simply change `T: Copy` to `T: Clone` as this will be inherently unsafe. The code operates based on the assumption that `T` cannot implement `Drop` (because `Copy` and `Drop` are mutually exclusive). Consequently, in `VarLenArray::<T>::drop()`, there's no inner destructors being called, the memory is just being freed at once. Additionally, just allowing `T: Clone` would allow you stuffing all kinds of Rust types into this which would make it even more unsafe.

One way to restrict it, I guess, would be to add something like
```rust
pub trait VarLenElement: 'static + Clone {}
impl<T: Copy> VarLenElement for T {}
impl<T: VarLenElement> VarLenElement for VarLenArray<T> {}
impl VarLenElement for VarLenString {}
```
and then require `T: VarLenElement` in `VarLenArray<T>`. In the destructor of `VarLenArray`, you'd have to manually drop all of the inner elements then.

And then it would need some tests to make sure it actually works.

PR #237, opened 2023-05-28
Avoid creating unaligned ref in H5Type derive macro

rytheo:
Creating unaligned references now causes a compile-time error, so this change uses `std::ptr::addr_of!` to compute struct field offsets instead.

### DIFF
--- a/hdf5-derive/src/lib.rs
+++ b/hdf5-derive/src/lib.rs
@@ -44,12 +44,16 @@ where
     F: ToTokens,
 {
     quote! {
-        let origin: *const #ty #ty_generics = ::std::ptr::null();
+        let origin = ::std::mem::MaybeUninit::<#ty #ty_generics>::uninit();
+        let origin_ptr = origin.as_ptr();
         let mut fields = vec![#(
             _h5::types::CompoundField {
                 name: #names.to_owned(),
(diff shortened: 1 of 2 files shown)


hdf5-derive/src/lib.rs
```diff
@@ -44,12 +44,15 @@ where
...
-        let origin: *const #ty #ty_generics = ::std::ptr::null();
+        let origin = ::std::mem::MaybeUninit::<#ty #ty_generics>::uninit();
+        let origin_ptr = origin.as_ptr();
         let mut fields = vec![#(
             _h5::types::CompoundField {
                 name: #names.to_owned(),
                 ty: <#types as _h5::types::H5Type>::type_descriptor(),
-                offset: unsafe { &((*origin).#fields) as *const _ as _ },
+                offset: unsafe {
+                    ::std::ptr::addr_of!((*origin_ptr).#fields) as usize - origin_ptr as usize
```
mulimoen, line 54: Maybe [`offset_from`](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset_from) with the pointers cast to u8 pointers?

mulimoen: Prefer `.cast::<u8>()` over `as`

mulimoen (approved):
Looks good, thanks!

PR #238, opened 2023-05-31
Apply lints

rytheo:
Fixes the latest clippy/fmt warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -236,7 +236,7 @@ jobs:
 
   msrv:
     name: Minimal Supported Rust Version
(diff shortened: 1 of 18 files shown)


mulimoen:
I'm wondering if this conflicts with MSRV. Could you bump the CI run for this test to ubuntu 20.04?

rytheo:
@mulimoen It does conflict. The current MSRV is 1.54; new format string syntax is 1.58 and let-else is 1.65.

rytheo:
Do we need to bump the MSRV or change the linter to not suggest new syntax?

PR #240, opened 2023-06-09
2023 fixups

aldanor:
Ok, this should allow to rebase the pending PRs and move forward with them, so we get green builds:

- Bump all dependencies at once (something may get broken, we'll see)
- Bump edition to 2021
- Bump MSRV to 1.64 (to use workspace inheritance)
- Use workspace inheritance for most crate metadata
- Use workspace inheritance for external/internal deps where possible
- Fix all accumulated clippy errors
- Migrate hdf5-derive to syn 2.0, migrate to the latest bitflags

Closes #238


aldanor:
@mulimoen I've noticed you mentioned in the other PR that ubuntu version should be bumped to `ubuntu 20.04`, wonder why? (should it be bumped in this PR as well then perhaps)

aldanor:
> Bitflags is dragged in by both us and `tempfile`, which wants an older version. This is just a lint and can be disabled by `-A clippy:multiple-crate-versions` (I think)

Oh, indeed. It's actually dragged by `rustix` which now has an [open PR](https://github.com/bytecodealliance/rustix/pull/687) that migrates it to bitflags 2.

aldanor:
Ok, so for now I think there's only one left - where on wine build it can't find the hdf5 library.

mulimoen:
I'd say we remove the wine check and open an issue on this. I can try to decipher the bug during the weekend

aldanor:
And now, out of a sudden, `msi @ 1.12` [fails](https://github.com/aldanor/hdf5-rust/actions/runs/5221174377/jobs/9425134193?pr=240)...

mulimoen:
Not that fun when the error goes away on a rerun.. We'll keep an eye on this.

Could all workspace related items (`[workspace]`, `[workspace.package]`, `[workspace.dependencies]`) be moved to the top of the toplevel Cargo.toml?

aldanor:
> Not that fun when the error goes away on a rerun.. We'll keep an eye on this.
Agreed, not fun at all :/

> Could all workspace related items (`[workspace]`, `[workspace.package]`, `[workspace.dependencies]`) be moved to the top of the toplevel Cargo.toml?
I wonder if we should just have a workspace-only Cargo.toml at the top?

And then `hdf5` crate just becomes a normal package and gets moved to `hdf5/` subfolder.

mulimoen:
> I wonder if we should just have a workspace-only Cargo.toml at the top?
> 
> And then `hdf5` crate just becomes a normal package and gets moved to `hdf5/` subfolder.

That would be very clean and nice

aldanor:
Ok, I've separate out the workspace and moved the main crate to `hdf5/` (also added you to crate authors if you don't mind), let's hope this doesn't break everything - if it builds green I'll merge it and we can then proceed with other PRs. (rebases may be a bit painful due to moving all these files around though...)

PR #243, opened 2023-06-09
Use 1.14.1 as bundled HDF5

mulimoen:
Replaces #207


mulimoen:
Hitting [some bug in hdf5](https://github.com/HDFGroup/hdf5/blob/ac07382ca947de2b936204dc9d60ce9ac245a21f/src/H5FDstdio.c#L354). Need to find a windows machine to find a fix and upstream this

aldanor:
Wonder if we could git-bisect it in a sane way...

mulimoen:
From `man off_t`
> off_t is used for describing file sizes.  It is a signed integer type.

Poorly defined in other words. There is a feature test macro we could set, `_FILE_OFFSET_BITS`, but I am not sure if this will affect other libraries

gauteh:
Would this PR give us a version with chunks_visit? 

* https://github.com/gauteh/hidefix/issues/30
* https://github.com/gauteh/hidefix/blob/main/src/idx/dataset/dataset.rs#L106C27-L106C27

mulimoen:
@gauteh Yes, this would provide `chunks_visit`.

gauteh:
Great! Looking forward to this!

mulimoen:
Still some mysterious errors here. On wine `StrStrIA` from `shlwapi` is not found (maybe this library has a different name on wine compared to windows?). `mingw` and `mingw-static` fails with `STATUS_STACK_BUFFER_OVERRUN` / `STATUS_HEAP_CORRUPTION`, although this seems to be the case in the current CI too.

hdf5-sys/build.rs
```diff
@@ -29,7 +29,7 @@ impl Version {
     }
 
     pub fn parse(s: &str) -> Option<Self> {
-        let re = Regex::new(r"^(1)\.(8|10|12|14)\.(\d\d?)(_\d+)?((-|.)(patch)?\d+)?$").ok()?;
+        let re = Regex::new(r"^(1)\.(8|10|12|14)\.(\d\d?)(_|.\d+)?((-|.)(patch)?\d+)?$").ok()?;
```
jedbrown, line 32: Is the `.` intended to be escaped `\.`? If not, the regex can be shortened. In any case, I now see version strings `"1.14.4-2"`.

mulimoen: Could probably simplify it greatly by ignoring everything after `number.number.number`

PR #245, opened 2023-06-10
Remove 1.13 support

mulimoen:
The HDF5 1.13 series was only a preview and requires some workarounds. Simplify by only supporting 1.14.

Adds 1.14.1 as supported since only a couple of new functions were added. These have not been included, but we are not aiming to cover all of HDF5 (already missing VFD etc.)

Fixed a small bug in `ChunkInfo`

### DIFF
--- a/hdf5-sys/build.rs
+++ b/hdf5-sys/build.rs
@@ -656,8 +656,8 @@ impl Config {
     pub fn emit_cfg_flags(&self) {
         let version = self.header.version;
         assert!(version >= Version::new(1, 8, 4), "required HDF5 version: >=1.8.4");
-        let mut vs: Vec<_> = (5..=21).map(|v| Version::new(1, 8, v)).collect(); // 1.8.[5-21]
-        vs.extend((0..=8).map(|v| Version::new(1, 10, v))); // 1.10.[0-8]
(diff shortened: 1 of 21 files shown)


hdf5-sys/build.rs
```diff
@@ -622,8 +622,7 @@ impl Config {
         let mut vs: Vec<_> = (5..=21).map(|v| Version::new(1, 8, v)).collect(); // 1.8.[5-21]
         vs.extend((0..=8).map(|v| Version::new(1, 10, v))); // 1.10.[0-8]
         vs.extend((0..=2).map(|v| Version::new(1, 12, v))); // 1.12.[0-2]
-        vs.extend((0..=0).map(|v| Version::new(1, 13, v))); // 1.13.[0-0]
```
aldanor, line 625: Should we update version checks while we're at it? The latest versions are: 1.8.23, 1.10.10, 1.12.2, 1.14.1.

mulimoen: I'll have a look at the changelog and update accordingly

mulimoen: No major changes in those patch versions

aldanor:
Could you rebase this on master just in case please? (and I think it should be go once it's green, always happy to reduce maintenance scope, so we won't see 1.13 ever again)

mulimoen:
Rebased and all green

aldanor (approved):

PR #246, opened 2023-06-10
Float16 support

### LINKED ISSUE #236: Slow performance compared to h5py
For my use case (large 2d table in h5 file, stored as f16 cast to f32 for use in rust), the `hdf5` package is significantly slower than taking a detour in python via `pyo3`. In python I use the `h5py` package, casting it to f32 using numpy functions and transferring the resulting array back into rust. Using only rust code (`hdf5`) takes up to 4.5 times as long as taking this detour. I assume that the issue is somewhere in the code that automatically casts values from one type to the other, since there are quite a few safety checks that could slow down the process, but I have not yet tested if the issue appears on native rust types (e.g. using raw f32 data as f32). I could imagine one of the following solutions:
- Reduce the amount of safety checks in the automatic casting code (probably not possible or the checks would not be there)
- Add optional unsafe functions to remove all checks when desired, similar to `get_unchecked` for `std::vec::Vec` or `uget` for `ndarray::ArrayBase`
- Enable the user to load data from the h5 file in its raw representation, i.e. return an uncast version of the array so that the user can perform whatever unsafe casting is desired, though I am not sure, how that would be possible in rust due to the uncertain return type. Maybe add generic raw-access functions that return an error whenever the generic does not match the raw type? In that case the user would need a getter for the raw type of the file, which could be used in a `match` statement.

All that is somewhat based on the premise, that casting truly is the issue, which I did not ensure, but the rest should be "just load the data from disk", so there is not much potential for slow-downs.

Below are two code snippets for loading a slice of a large table via `pyo3`+`h5py` and `hdf5`, respectively.

<details>
<summary>pyo3+h5py</summary>

```rust
// Trait to access the numpy name of rust primitives
pub trait NumpyEquivalent: numpy::Element {
    fn numpy_name() -> &'static str;
}
macro_rules! make_numpy_equivalent {
    ($(($rust_types: ty, $numpy_names: literal)),*) => { $(make_numpy_equivalent!($rust_types, $numpy_names);)* };
    ($rust_type: ty, $numpy_name: literal) => {
        impl NumpyEquivalent for $rust_type {
            fn numpy_name() -> &'static str { $numpy_name }
        }
    };
}
make_numpy_equivalent!((f32, "float32"), (f64, "float64"), (bool, "bool_"), (u8, "uint8"), (u16, "uint16"), (u32, "uint32"), (u64, "uint64"));
// Function to access a slice of a h5 file in a desired primitive type
fn get_rows_slice_pyo3<T: NumpyEquivalent>(file: &str, dataset: &str, i_row_from: usize, i_row_to: usize) -> Result<ndarray::Array2<T>, pyo3::PyErr> {
    pyo3::Python::with_gil(|py| {
        let locals = pyo3::types::PyDict::new(py);
        locals.set_item("h5py", py.import("h5py")?)?;
        locals.set_item("np", py.import("numpy")?)?;
        locals.set_item("data", py.eval(
            format!("h5py.File(\"{:}\")[\"{:}\"]", file, dataset).as_str(),
            None,
            Some(&locals)
        )?)?;
        locals.set_item("start", i_row_from)?;
        locals.set_item("end", i_row_to)?;
        let row_obj = py.eval(
            format!("data[start:end].astype(np.{:})", T::numpy_name()).as_str(),
            None,
            Some(&locals)
        )?;
        let row: &numpy::PyArray2<T> = row_obj.downcast()?;
        Ok(row.to_owned_array())
    })
}
```
</details>

<details>
<summary>hdf5</summary>

```rust
fn get_rows_slice_hdf5<T: hdf5::H5Type>(file: &str, dataset: &str, i_row_from: usize, i_row_to: usize) -> Result<ndarray::Array2<T>, hdf5::Error> {
    let file = hdf5::File::open(file)?;
    let data = file.dataset(dataset)?;
    data.read_slice_2d(ndarray::s![i_row_from..i_row_to,..])
}
```
</details>

<details>
<summary>Necessary Cargo.toml lines</summary>

```toml
[dependencies]
ndarray = "0.15.6"
hdf5 = "0.8.1"
pyo3 = {version="0.18.3", features=["auto-initialize"]}
```
</details>

<details>
<summary>Software versions</summary>

- hdf5: 0.8.1
- libhdf5-dev: 1.10.8
- pyo3: 0.18.3
- ndarray: 0.15.6
- python: 3.11.2
- numpy: 1.24.3
- h5py: 3.8.0
</details>

aldanor:
Closes #236 (?)


aldanor:
- Both `f16` and `complex` features need at least one tiny test perhaps, just to check that a single number can roundtrip from rust -> hdf5 -> rust?... (@mulimoen maybe you could try that in a separate PR?)
- Will probably also refactor DynValue in this PR as well, packing floats into one variant, now that there's (up to) 3 of them
- "complex" and "f16" are now top-level features in the main crate

aldanor:
@mulimoen In the failures, your `test_issue_223` test has failed apparently 🤔

(btw I moved the tests under `hdf5/` now, so previously things might have been running slightly incorrectly...)

.github/workflows/ci.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       - name: Document workspace
         env:
           RUSTDOCFLAGS: "--cfg docsrs"
-        run: cargo doc --features static,zlib,blosc,lzf
+        run: cargo doc --features static,zlib,blosc,lzf,f16,complex
```
mulimoen, line 49: These flags should be added to `package.metadata.docs.rs` in `Cargo.toml`, both for the `hdf5` package and the `hdf5-types` (which lacks this entry).

hdf5-types/src/dyn_value.rs
```diff
@@ -115,19 +115,76 @@ impl From<DynInteger> for DynValue<'_> {
 }
 
 #[derive(Copy, Clone, PartialEq)]
-pub enum DynScalar {
-    Integer(DynInteger),
+pub enum DynFloat {
```
mulimoen, line 118: Bit of an edge case but we could add `#[cfg_attr(not(feature = "f16"), non_exhaustive)]
`

hdf5-types/src/h5type.rs
```diff
@@ -40,18 +40,24 @@ impl IntSize {
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
```
mulimoen, line 41: Same here with `#[cfg_attr(not(feature = "f16"), non_exhaustive)]`

mulimoen (approved):
With the small nits and #247 merged this looks great!

hdf5-types/src/h5type.rs
aldanor: Tried it out, but how would you do this then: https://github.com/aldanor/hdf5-rust/blob/46f2bccc30e1f440484e9fcf3bb5e35e6804de05/hdf5/src/hl/datatype.rs#L354-L359

mulimoen: Yeah, it seems bad having to panic there. Very fringe case so skip my suggestion

.github/workflows/ci.yml
aldanor: Yea, good point - added them to both.

aldanor:
Enabled f16/complex for docs.rs builds for both hdf5 and hdf5-types - thanks, good call.

As you said above, exhaustive attr might not be the best idea since it would force ourselves to panic when matching which is a bit weird.

If it builds green, I think it should be good to go :)

mulimoen:
Looks good to go if green!

PR #247, opened 2023-06-11
Only run test_issue_223 when deflate filter is avail

mulimoen:
A meta-PR targeting the branch of #246

### DIFF
--- a/hdf5-sys/build.rs
+++ b/hdf5-sys/build.rs
@@ -175,8 +175,8 @@ pub struct Header {
     pub have_direct: bool,
     pub have_parallel: bool,
     pub have_threadsafe: bool,
-    pub have_zlib: bool,
     pub have_no_deprecated: bool,
+    pub have_filter_deflate: bool,
     pub version: Version,
 }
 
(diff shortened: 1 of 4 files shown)


hdf5-sys/build.rs
```diff
@@ -204,6 +205,7 @@ impl Header {
                 hdr.have_threadsafe = value > 0;
             } else if name == "H5_HAVE_FILTER_DEFLATE" {
                 hdr.have_zlib = value > 0;
+                hdr.have_filter_deflate = value > 0;
```
aldanor, line 208: Ok, I'm somewhat confused, in HDF5 terminology isn't "deflate" = "zlib"?

mulimoen: They are the mostly the same and used interchangeably. The compile definition is `HDF5_ENABLE_Z_LIB_SUPPORT` but the flag in `H5pubconf.h` is `H5_HAVE_FILTER_DEFLATE`.

mulimoen: I don't mind having just `have_zlib` or `have_filter_deflate`

aldanor: Hm, just a single `have_filter_deflate` is probably more correct since that's the flag we're checking in `H5pubconf`?

aldanor:
👍 I'll merge it into the f16 branch and check the builds there

PR #249, opened 2023-06-24
Add minimal documentation for public items

rytheo:
Related to #141


aldanor:
Btw, on a related note, one of the problems we had was this: https://github.com/rust-lang/rust/issues/32077 (@mulimoen)

However, it was actually fixed a few weeks ago!
// but then reverted because of a bug
// // but reverting will be reverted again soon after the bug is fixed

aldanor:
@rytheo Thank you very much for contributing, I'll get back to you with a review and a few minor comments.

PR #254, opened 2023-08-10
Mark various internal functions as unsafe

rytheo:
Found various functions that should be `unsafe` since they take raw pointer arguments and must assume they point to valid data.

Made `string_from_fixed_bytes` panic on invalid UTF-8 instead of marking it unsafe, but marking it unsafe is a valid alternative if the `unwrap` call significantly reduces performance.

### DIFF
--- a/hdf5-sys/src/h5.rs
+++ b/hdf5-sys/src/h5.rs
@@ -117,7 +117,7 @@ extern "C" {
 }
 
 #[cfg(feature = "1.14.0")]
-type H5_atclose_func_t = Option<extern "C" fn(ctx: *mut c_void)>;
+type H5_atclose_func_t = Option<unsafe extern "C" fn(ctx: *mut c_void)>;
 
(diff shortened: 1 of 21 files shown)


mulimoen (approved):
Looks good, this will allow storing C functions directly in the type without wrapping it up in a "safe" function.

Marking the functions in `hdf5/src/util.rs` as `unsafe` is a breaking change. Could this be added to the changelog?

rytheo:
@mulimoen changing `hdf5/src/util.rs` shouldn't be a breaking change; looking at `hdf5/src/lib.rs`, it seems like none of the functions in `util` are accessible outside the crate. Am I missing something?

mulimoen:
@rytheo You are right, after closer examination it is not exported publicly. Note to self: Should mark these functions/modules as `pub(crate)`

aldanor (approved):

PR #260, opened 2023-09-26
Only conditionally call H5Pset_evict_on_close

### LINKED ISSUE #259: Call to `H5Pset_evict_on_close` cannot be prevented and panics in parallel builds
I was experimenting with parallel hdf5 by using the mpio feature gate. Now I wasn't exactly sure about how to do this but came up with the following program:

```
use hdf5::{
    file::LibraryVersion,
    plist::{self},
    FileBuilder,
};
use mpi::traits::{AsRaw, Communicator};

const NUM_ELEMENTS: usize = 10000000;
const FILE_NAME: &str = "out.hdf5";
const DATASET_NAME: &str = "data";

fn main() {
    let universe = mpi::initialize().unwrap();
    let rank = universe.world().rank() as usize;
    let num_ranks = universe.world().size() as usize;

    let comm = universe.world();
    let fapl = plist::FileAccess::build()
        .mpio(comm.as_raw(), None)
        .libver_bounds(LibraryVersion::V18, LibraryVersion::V110)
        .finish()
        .unwrap();
    let fcpl = plist::FileCreate::build().finish().unwrap();
    let f = FileBuilder::new()
        .set_access_plist(&fapl)
        .unwrap()
        .set_create_plist(&fcpl)
        .unwrap()
        .create(FILE_NAME)
        .unwrap();
    let dataset = f
        .new_dataset::<f64>()
        .shape(&[NUM_ELEMENTS * num_ranks])
        .create(DATASET_NAME)
        .unwrap();
    let values: Vec<_> = (0..NUM_ELEMENTS).map(|_| rank as f64).collect();
    dataset
        .write_slice(&values, rank * NUM_ELEMENTS..(rank + 1) * (NUM_ELEMENTS))
        .unwrap();
}
```

If I compile this using hdf5-openmpi and run it (with the current master branch on commit `26046fb`), I get
`H5Pset_evict_on_close(): evict on close is currently not supported in parallel HDF5`
during the call to `FileBuilder::create`.

I initially thought I would just call `FileAccess::build().evict_on_close(false) ...` for the file access list, but this doesn't work because the `FileAccessBuilder` internally stores `Some(false)` instead of `None`, and `H5Pset_evict_on_close` is called either way in `populate_plist` due to the following code:
```
if let Some(v) = self.evict_on_close {
    h5try!(H5Pset_evict_on_close(id, hbool_t::from(v)));
}
```

The `FileAccessBuilder` is initially created with `evict_on_close: None` but during `FileAccessBuilder::from_plist`, `evict_on_close` is set to `Some(...)`:
```
            builder.evict_on_close(plist.get_evict_on_close()?);
```
I didn't see any way to internally set `evict_on_close` to None before file creation, so that I always end up with this panic.

Now I managed to get this program running by using a local version of hdf5 in which I commented out the line above. Is there any way that this could be fixed officially?

mulimoen:
Closes #259

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -273,7 +273,7 @@ jobs:
         with: {submodules: true}
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
-        with: {toolchain: 1.64}
+        with: {toolchain: "1.70"}
       - name: Build and test all crates
         run:
           cargo test --workspace -vv --features=hdf5-sys/static,hdf5-sys/zlib --exclude=hdf5-derive
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 - Applying filters without chunking will now produce an explicit error.
 - Fixed a bug where chunking could not be enabled for zero-sized extents.
 - Fixed library finding on Windows with MSYS2-distributed MinGW HDF5.
+- Fixed a bug which made parallel builds unusable.
 
 ## 0.8.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 ### Changed
 
 - The `H5Type` derive macro now uses `proc-macro-error` to emit error messages.
-- MSRV is now `1.64.0` and Rust edition has now been bumped to 2021.
+- MSRV is now `1.70.0` and Rust edition has now been bumped to 2021.
 - Types in ChunkInfo has been changed to match HDF5.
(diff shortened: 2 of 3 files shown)


aldanor:
Looks like MSRV tests fail because the most recent `half` (required by `feature="f16"`) bumped their MSRV to 1.70 🤔 Wonder if we should do the same then

mulimoen:
Alternatively make MSRV only work for the minimal set of features, but I'd rather bump it up to 1.70 even with 1.70 being quite recent (1st of June)

aldanor (approved):

PR #261, opened 2023-11-08
Fix definition of `H5L_info2_t`

soph-dec:
Replace incorrect `H5L_info1_t__u` with `H5L_info2_t__u`. Since `H5L_info1_t__u` does not have the same size as `H5L_info2_t__u`, this bug resulted in 8 bytes of stack memory outside the allocated region being overwritten by the hdf5 C-library for example when using `H5Lget_info2`.

Since `Debug` can no longer be derived, add a manual implementation for the struct `H5L_info2_t` in order to correctly match the variant of `H5L_info2_t__u` to the link type.

Additionally, add methods to access the union fields of `H5L_info2_t__u`.

### DIFF
--- a/hdf5-sys/src/h5l.rs
+++ b/hdf5-sys/src/h5l.rs
@@ -1,4 +1,5 @@
 //! Creating and manipulating links within an HDF5 group
+#[cfg(feature = "1.12.0")]
 use std::fmt;
 use std::mem;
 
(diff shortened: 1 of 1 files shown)


mulimoen (approved):
Great catch in finding this bug

soph-dec:
I pushed one more commit to fix the clippy error in hdf5-sys. Some of the other tests are still failing, but I do not know how this could be connected to my changes.

mulimoen:
The clippy and MSRV jobs are known to be failing. On windows we have started to get some random failures, although unrelated to this change

soph-dec:
I see, thanks for clarifying!

PR #268, opened 2024-01-30
Maintenance updates

aldanor:
Some long overdue maintenance updates.

(We should prob release a version soon, are we still stuck with the 1.14 problem?..)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -293,7 +293,11 @@ jobs:
       - name: Build and test
         env:
           CARGO_TARGET_X86_64_PC_WINDOWS_GNU_RUNNER: wine64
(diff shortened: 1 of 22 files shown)


mulimoen (approved):
Going to rubber stamp this, would be great to have a new version of these crates out.

I think the `wine` failure is due to `hdf5` built as `dev` does not build from source with the new resolver version. One could `--exclude hdf5-derive` for this step, and add another step with `cargo test --features hdf5-derive --feature hdf5/static`. Update: This fixed compilation, unfortunately running the compile time tests requires some additional magic. I think we can disable this test for wine, as we do for other jobs in CI

The `mingw` failure I have not been able to replicate locally

Issue #98, opened 2020-06-11
Request: Reference types


gauteh:
Hi,

while figuring out how to get the dimensions of a dataset I am trying to access  `STD_REF_OBJ`'s in attributes (using #65 with changes from @balintbalazs). A simple example of references are generated with this script:

```py
from h5py import File
import numpy as np

# http://docs.h5py.org/en/stable/high/dims.html

f = File('dims_1d.h5', 'w')

f['x1'] = [1, 2]
f['x1'].make_scale('x1 name')
f['data'] = np.ones((2,), 'f')
f['data'].dims[0].attach_scale(f['x1'])
```

which gives the following structure (using `h5dump -A dims_1d.py`):

```
HDF5 "dims_1d.h5" {
GROUP "/" {
   DATASET "data" {
      DATATYPE  H5T_IEEE_F32LE
      DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
      ATTRIBUTE "DIMENSION_LIST" {
         DATATYPE  H5T_VLEN { H5T_REFERENCE { H5T_STD_REF_OBJECT }}
         DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
         DATA {
         (0): (DATASET 800 /x1 )
         }
      }
   }
   DATASET "x1" {
      DATATYPE  H5T_STD_I64LE
      DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
      ATTRIBUTE "CLASS" {
         DATATYPE  H5T_STRING {
            STRSIZE 16;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
         DATA {
         (0): "DIMENSION_SCALE"
         }
      }
      ATTRIBUTE "NAME" {
         DATATYPE  H5T_STRING {
            STRSIZE 8;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
         DATA {
         (0): "x1 name"
         }
      }
      ATTRIBUTE "REFERENCE_LIST" {
         DATATYPE  H5T_COMPOUND {
            H5T_REFERENCE { H5T_STD_REF_OBJECT } "dataset";
            H5T_STD_I32LE "dimension";
         }
[...]

PR #279, opened 2024-04-13
Minimal Object Reference Implementation

JamesMc86:
This PR includes changes to support object references as requested in #98.

It adds ObjectReference1 as pre-1.12 object references and ObjectReference2 as the new reference types based on the STD_REF type.

I've tried to make this useful without being too large and can build on it further.

Key questions for me are around the terminology as I'm quite new to HDF5 - let me know if that makes sense.

Other possible issues to consider:
* Could we store the location in the references as well as it gets passed on nearly every call?
* There is a todo in the DynValue that @mulimoen included - I'm not familiar with this to know what to do but I assume a supported type needs adding before release.
* ObjectReference2 can't be supported in VarArray due to not implementing a Copy. I would propose adding a TryClone type that VarArray can take so we can support this. I'm happy to do this in another release.
* The same basic structure can be used to add Region references and Attribute references in a future version.


JamesMc86:
Obviously got into a bit of a mess here still so will mark as draft. I thought I had it.

One problem is bugs in 1.12.0 with references so now flagged as 1.12.1 required.

Another is with the VarLenArray I believe. Going to back out the changes to make them work with Non-Copy types and the ObjectReference2 will just not be supported in them at this point.

I've successfully got conda replicating another issue on 1.8 so I'll work on that too!

hdf5-types/src/dyn_value.rs
```diff
@@ -703,6 +703,7 @@ impl<'a> DynValue<'a> {
             FixedUnicode(_) => DynFixedString::new(buf, true).into(),
             VarLenAscii => DynVarLenString::new(buf, false).into(),
             VarLenUnicode => DynVarLenString::new(buf, true).into(),
+            Reference(_x) => todo!(),
```
mulimoen, line 706: Should we support this?

hdf5/src/hl/references/legacy.rs
```diff
@@ -0,0 +1,69 @@
...
+#[cfg(not(feature = "1.12.0"))]
+use hdf5_sys::h5r::H5R_OBJECT as H5R_OBJECT1;
+#[cfg(feature = "1.12.0")]
+use hdf5_sys::h5r::H5R_OBJECT1;
+
+use crate::{Location, ObjectReference};
+
+#[repr(transparent)]
+#[derive(Debug, Copy, Clone)]
+pub struct ObjectReference1 {
```
mulimoen, line 21: Move to `hdf5-types`

hdf5/src/hl/references/legacy.rs
```diff
@@ -0,0 +1,69 @@
...
+
+#[cfg(not(feature = "1.12.0"))]
+use hdf5_sys::h5r::H5R_OBJECT as H5R_OBJECT1;
+#[cfg(feature = "1.12.0")]
+use hdf5_sys::h5r::H5R_OBJECT1;
+
+use crate::{Location, ObjectReference};
+
+#[repr(transparent)]
+#[derive(Debug, Copy, Clone)]
```
mulimoen, line 20: I don't think you can implement `Copy` and `Clone` for these?

hdf5/src/hl/references/standard.rs
```diff
@@ -0,0 +1,104 @@
...
+use hdf5_sys::h5r::H5R_type_t::H5R_OBJECT2;
+use hdf5_sys::h5r::{H5R_ref_t, H5Rcreate_object, H5Rdestroy, H5Rget_obj_type3, H5Ropen_object};
+
+use super::ObjectReference;
+use crate::internal_prelude::*;
+use crate::Location;
+
+/// A reference to a HDF5 item that can be stored in attributes or datasets.
+#[repr(transparent)]
+pub struct StdReference(H5R_ref_t);
```
mulimoen, line 15: Move to `hdf5-types`

mulimoen (changes requested):
This looks great, just move a couple of structs and this should be ready to merge.

mulimoen:
I think we can lift the restriction of `Copy` on `H5Type` if we require derived types to be `Copy`. But this depends on what sort of issues we got for `VarLenArray<ref>` which could suggest something worse going on.

hdf5/src/hl/references/legacy.rs
JamesMc86: I was a little unclear on this. I went to remove it but these references aren't actually supported by H5Rdestroy or H5Rcopy. I'm not sure how to test for this in the absence of explicit documentation. I'll take a look at the HDF5 source for any clues

mulimoen: Could be helpful to have a look in the header files of a 1.8 release

JamesMc86:
> I think we can lift the restriction of `Copy` on `H5Type` if we require derived types to be `Copy`. But this depends on what sort of issues we got for `VarLenArray<ref>` which could suggest something worse going on.

Yes I think there is a way - the main issue is the ptr lifetime and implementing our own drop routine. I've been trying to think if there is a way to treat it as a slice so we get some for free but I guess it will always have the problem of needing to be C compatible with the HDF5 library.

I'm happy to have a go with this as well but didn't want to grow this PR any bigger!

hdf5/src/hl/references/legacy.rs
JamesMc86: Ah perhaps these should use H5O_close - I'll give this a go

JamesMc86: I've done a bit of digging and I think these can be copy but let me know if you think this is wrong:

* Found this example which shows them freed as normal memory: https://github.com/HDFGroup/hdf5/blob/1d1c0405449b70656d9ec30afa95d358fc2719a0/examples/h5_reference_deprec.c#L82
* The 1.8 docs show that it `hobj_ref_t` is an typdef for `haddr_t` which is a 64 bit address in the file.

hdf5/src/hl/references/standard.rs
JamesMc86: Attempted this move but we need access to h5call! for the function in the drop implementation.

hdf5/src/hl/references/legacy.rs
JamesMc86: I have a version with this moved but if we can't move StdReference does it make sense to keep it where it is for consistency?

mulimoen: I am OK keeping it here if we can't move `StdReference`

hdf5-types/src/dyn_value.rs
mulimoen, line 706:
```suggestion
            Reference(_x) => todo!("References are not yet supported as fill value"),
```

mulimoen: Since we can't use `h5call` and family in `hdf5-types` we will struggle to support this feature without some heavy refactoring.

mulimoen (approved):

JamesMc86:
Just wanted to check what was blocking this and #243 as they will start blocking my project. Happy to do work here to unblock it. I see the falling brew install was removed in another PR anyway. I can try and triage the MinGW issue too

mulimoen:
I would like an approval from @aldanor, but for me this is looking good to go. Note that only @aldanor has rights to release as of now (and a release is long overdue).

It would be great if you managed to track down the mingw error, it is present also on `master` so could be a separate PR.

PR #284, opened 2024-05-23
fix: Add Installation to Conda setup in CI

JamesMc86:
It seems from https://github.com/actions/runner-images/issues/9262 that miniconda was removed from Mac images.

This adds the configuration to the setup step to now install miniconda again.

I need to validate that this isn't having a significant impact on the Windows and Linux steps where this is not a concern.


JamesMc86:
Disregard this for now - I see the bigger problem is that MacOs-latest is now ARM based which causes other issues.

I'll create an issue to discuss the best solution.

Issue #286, opened 2024-05-23
Fix Wine CI Runner


JamesMc86:
Not got time to look at this now but thought I would share my notes.

I noticed the Wine runner is now broken. This is related to Rust 1.76 dropping support for Windows 7 and now requires the bcryptprimitives.dll.

From research, this was added in Wine 8.13, but the version in the Ubuntu repos is much older, so the CI system would need to install Wine from a newer repo, like adding the Wine repos.

mulimoen:
I suppose pinning to rust `1.75` is a pragmatic fix?

PR #289, opened 2024-05-30
fix: Unavailable Versions in CI

JamesMc86:
* Remove HDF5 1.8 from the brew tests - no longer available.
* Limit Wine tests to Rust 1.75 to maintain compatilibity with the installed wine version. #286


mulimoen (approved):

mulimoen:
Thanks!

